PR #8, opened 2022-03-23
CollisionObject

samjgorman:
### Description of Changes
Introducing a moderately extensible CollisionObject that uses collision detection from the cannon.js physics engine.
Position, size, callback upon collision, mass may be modified through props.  Future work will add more shapes to the CollisionObject component, or simply use this as a template for future components that use collision detection. 

image
Standing on CollisionObject for reference.

### Link to Notion ticket
This work touches multiple tickets, but is most closely related to setting world boundaries. 

### Testing Procedure
Tested locally in browser.


samjgorman:
@kanyesthaker  @shalins ready for review!

kanyesthaker:
Nice work! This is a great starting point for object collision, definitely one of the more difficult yet important components of this project.

I think I'd prefer if the collision logic were a little more malleable before merging into `main`. For example, it would be neat to more cleanly attach collision to any object (glTF or native Three.js). I think it could be neat to aim for a pure TS function that allows us to define collision on _any_ (within reason) object, versus needing to modify a template (since a lot of our objects will likely be so complex/differentiated that a template-based approach may not be convenient).

You may have already seen examples of using React hooks for Cannon.js with glTFs over at https://cannon.pmnd.rs/ with the source code for these examples [here](https://github.com/pmndrs/use-cannon/tree/master/packages/react-three-cannon-examples) -- would be neat if we could abstract away needing to explicitly call `useSphere` or `useBox` on every object in a scene, and instead automatically apply collision to them!

samjgorman:
I'm closing this PR and opening a new PR that describes the new, more robust approach.

PR #11, opened 2022-03-30
Upgrade node dependencies to resolve security vulnerabilities

kanyesthaker:
### Description of Changes
Re-install node-modules to resolve security vulnerabilities with nth-check that are unable to be fixed by dependabot

### Link to Notion ticket
N/A

### Testing Procedure
N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,25 +64,25 @@
       }
     },
     "node_modules/@babel/compat-data": {
-      "version": "7.17.0",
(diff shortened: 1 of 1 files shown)


shalins (approved):
This is an ENORMOUS diff but it seems like it's just upgrading a whole bunch of packages and their dependencies. And since it's not actually making changes to our codebase, it doesn't require a rigorous review.

PR #12, opened 2022-04-09
Collisions with shape library 

samjgorman:
### Description of Changes
Introduce a shape library of components that hook into Cannon's physics engine and recognize collisions.

1. introduces box, cylinder, and sphere components that accept props that make them extensible, such as defining a onCollision callback, defining visual properties such as color and size, and its position in the world.  Future work may easily introduce new shapes as needed to the shapes library.

2. Introduces a GLTFCollisionModel component in the shapes library that provides an abstraction for loading and defining collisions for GLTFs.  Under the hood, it uses GLTFLoader to load the model, then constructs a computationally inexpensive bounding box based on the geometry of the GLTF. Rendering collisions with a simple box mesh rather than a complex convex mesh based on the GLTF geometry has positive performance implications and better enables scaling the 3D world up to include multiple GLTF objects with higher numbers of vertices & edges.  This component is extensible and accepts props that scale the model size & bounding box size, along with the default props of other shapes in the shape library. 

Screen Shot 2022-04-09 at 12 27 27

### Link to Notion ticket
This unblocks multiple tickets but is not explicitly written as a ticket currently. 
Unblocked tickets: 
https://www.notion.so/irisapp/Add-sloped-surfaces-with-the-appropriate-collision-physics-b70c54e8c9184bb0bfb4238a835eefae

### Testing Procedure
Tested in browser with a variety of GLTF objects of different sizes & scales.  Default shapes in shapes library were individually tested.


samjgorman:
@kanyesthaker @shalins ready for review!

src/Player/playerMovementControls.tsx
```diff
@@ -29,22 +29,21 @@ export const playerMovementControls = () => {
...
         [keyMap[key]]: true,
       }));
     };
     const onKeyUp = (event: { key: string }) => {
       let key = event.key;
-      if (event.key == ' '){
+      if (event.key == ' ') {
         key = 'Space';
       }
-      setDirection(pressedKey => ({
```
shalins, line 47: these all seem like linter changes -- can we use a linter to check for these so they're not in the code reviews?

src/Player/playerMovementEmitter.ts
```diff
@@ -12,6 +12,7 @@ import movementLog from './playerMovementLog';
 let playerMovementEmitter = new EventEmitter();
 playerMovementEmitter.on('sendCoords', async (data: PlayerPosition) => {
   const packagedData = serialize(data);
+
```
shalins, line 15: Again, things like this shouldn't be in the PR. It seems like a lot of the code that was merged in was not linted. Since linting generally makes code reviews more obfuscated, they should not be added. 

Instead, we should make a separate PR just for linting and merge that in main.

@kanyesthaker you added an eslint, correct? Any steps @samjgorman  and I need to take to make sure the code is consistent?

src/Scene/Scene.tsx
```diff
@@ -4,11 +4,12 @@ import { Canvas } from '@react-three/fiber';
...
 import Floor from './Floor';
 import PlayerCameraControls from '../Player/playerCameraControls';
-import SceneObject from './SceneObject';
 import Skybox from './Skybox';
-import GLTFModel from './GLTFModel';
+import Box from '../Shapes/Box';
+import Sphere from '../Shapes/Sphere';
 import { getS3Path } from '../utils';
-
+import Cylinder from '../Shapes/Cylinder';
```
shalins, line 11: These are seem like they should be grouped. For better readability, can we have it be one import with the modules that are useful? 

`import { Box, Cylinder, Sphere } from '../Shapes'`

shalins, line 11: We should follow a convention for imports, because they can quickly grow out of proportion and start detracting from code readability. At Matician, we have simple enforcements such as organizing imports alphabetically, placing stdlib imports before 3rd party imports, etc. 

I don't have strong preference what the import convention should be for us, but I do think we should maintain one. For example, ESLint has their own sets of rules. In general, I think all hooks and react native imports should go at the top, `default` imports should go next, and `named` imports should go last.

As an example, something like:

```
import React from 'react';
import ReactDOM from 'react-dom';

import Floor from './Floor';
import PlayerCameraControls from '../Player/playerCameraControls';

import { Box, Cylinder, Sphere } from '../Shapes'
import { getS3Path } from '../utils';
```

Again, I think a linter can be configured to take care of most of these things. See: https://stackoverflow.com/a/61208572/1376311

src/Scene/Scene.tsx
```diff
@@ -18,17 +19,106 @@ export default function Scene() {
...
-            <SceneObject/>
-            <GLTFModel position={[10, 11, 60]} modelPath={getS3Path('models/gunship/scene.gltf')} scaleFactor={0.03}/>
-            <Floor/>
-            <gridHelper args={[100, 100, 'black', 'grey']}/>
-          </Physics>
+        <PlayerCameraControls />
+        <Skybox />
+        <ambientLight />
+        <pointLight position={[10, 10, 10]} />
+        <Physics gravity={[0, -30, 0]}>
```
shalins, line 26: What are these magic numbers? They should be given very descriptive names and added to a constants file where they can easily be tuned by anyone. Then import and use them here.

src/Scene/Scene.tsx
```diff
@@ -18,17 +19,106 @@ export default function Scene() {
...
+          <Player />
+          <Sphere
+            args={[10]}
+            mass={10}
+            position={[5, 1, -15]}
+            type={'Static'}
+            onCollide={() => console.log('I collided!')}
```
shalins, line 33: Don't merge a console log into main. Please add a better placeholder here and write a `TODO(SAM): description of what to do"` or `FIXME(SAM): description of what to fix"` here so it's clear to anyone reading or editing the code that this functionality has not been implemented yet.

src/Scene/Scene.tsx
```diff
@@ -18,17 +19,106 @@ export default function Scene() {
...
+            color={'red'}
+          />
+
+          <Box
+            args={[10, 10, 20]}
```
shalins, line 38: More magic numbers. Additionally, can we have a more descriptive name than `args`? I like how we have `mass` and `position`, however it's not clear what `args` is.

src/Scene/Scene.tsx
```diff
@@ -18,17 +19,106 @@ export default function Scene() {
...
+            mass={10}
+            position={[25, 1, -15]}
+            type={'Static'}
+            onCollide={() => console.log('I also collided!')}
+            color={'yellow'}
```
shalins, line 43: Even though this is not a magic number, we should create an object for all the standard colors so that we're not passing strings. The use of strings is quite brittle as a simple `grey` vs `gray` could cause unintended behavior.

src/Scene/Scene.tsx
```diff
@@ -18,17 +19,106 @@ export default function Scene() {
...
+        <Skybox />
+        <ambientLight />
+        <pointLight position={[10, 10, 10]} />
+        <Physics gravity={[0, -30, 0]}>
+          <Player />
+          <Sphere
+            args={[10]}
+            mass={10}
+            position={[5, 1, -15]}
+            type={'Static'}
```
shalins, line 32: No idea what this means. Can we make this more descriptive and avoid passing strings?

src/Scene/Scene.tsx
```diff
@@ -18,17 +19,106 @@ export default function Scene() {
...
+          />
+
+      <GLTFCollisionModel
+          mass={10}
+          position={[-50, 30, -50]}
+          type={'Static'}
+          onCollide={() => console.log('I also collided!')}
+          modelScaleFactor={0.001}
+          bboxScaleFactor={0.5}
+          modelPath={getS3Path('models/gunship/scene.gltf')}
```
shalins, line 53: Can we not hardcode these paths?

src/Shapes/Sphere.tsx
```diff
@@ -0,0 +1,31 @@
...
+interface CustomSphere extends SphereProps{
+  color: string
+}
+/**
+ * Sphere component renders a sphere with collision detection.
+ * This component accepts props that determine its size, position, type, and mass,
+ * and a callback to be called upon a collision.
+ * @param props args: Array<number>, mass: number, position: Triplet, type: string, onCollide: Function
+ * @returns
+ */
```
shalins, line 13: Nice work on documentation! This was very helpful :)

src/Scene/Scene.tsx
```diff
@@ -18,17 +19,106 @@ export default function Scene() {
...
+          /> 
+
+      <GLTFCollisionModel
+          mass={10}
+          position={[50, 0, 30]}
+          type={'Static'}
+          onCollide={() => console.log('I also collided!')}
+          modelScaleFactor={1}
+          bboxScaleFactor={0.5}
+          modelPath={getS3Path('models/plant2/scene.gltf')}
```
shalins, line 82: Instead of calling `getS3Path` in here, can we call these when defining the constants and just use a variable here instead? It would help separate out the logic.

For example, something like:
```
const tropicalPlant = getS3Path('models/plant/scene.gltf');
const desertPlant   = getS3Path('models/plant2/scene.gltf');
```

and then use it as:
```
...
          modelPath={Plants.desertPlant}
...
```

src/Scene/Scene.tsx
```diff
@@ -18,17 +19,106 @@ export default function Scene() {
...
+          />
+
+          <GLTFCollisionModel
+          mass={10}
+          position={[5, 0, 50]}
+          type={'Static'}
+          onCollide={() => console.log('I also collided!')}
+          modelScaleFactor={1}
+          bboxScaleFactor={1}
+          modelPath={getS3Path('models/plant/scene.gltf')}
```
shalins, line 72: Nit: In s3, can we give these more descriptive file paths? What is the difference between plant and plant2? Generally, the less unneeded obfuscation there is the better.

src/Scene/Scene.tsx
```diff
@@ -18,17 +19,106 @@ export default function Scene() {
...
+          type={'Static'}
+          onCollide={() => console.log('I also collided!')}
+          modelScaleFactor={0.3}
+          bboxScaleFactor={0.5}
+          modelPath={getS3Path('models/mushroom_tree/scene.gltf')}
+          />  
+
+
+          <Floor />
+          <gridHelper args={[100, 100, 'black', 'grey']} />
```
shalins, line 120: 🙂  Here you used `grey` with an `e` whereas I spell it with an `a`, which is why I made the point to namespace these.

src/Scene/SceneObject.tsx
```diff
@@ -11,11 +11,16 @@ export default function SceneObject() {
...
+      setMatrixAt: any;
+    }>();
 
   useFrame(({ clock }) => {
     const time = clock.getElapsedTime();
-    if (!meshRef.current){
+    if (!meshRef.current) {
       return;
     }
     if (meshRef.current) {
```
shalins, line 26: Can we turn this into an `if-else` block and add a comment why we're returning early?

src/Scene/SceneObject.tsx
```diff
@@ -46,10 +51,9 @@ export default function SceneObject() {
   });
 
   return (
-    <instancedMesh ref={meshRef} args={[null, null, 1]} position={[0, 20, 0]}>
+    <instancedMesh ref={meshRef} args={[null, null, 1]} position={[0, 5, 0]}>
```
shalins, line 54: Again, since AFAIK, these are magic numbers I have no idea how to interpret this change.

src/Shapes/Box.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { BoxProps, useBox } from '@react-three/cannon';
+
+interface CustomBox extends BoxProps {
+  color: string;
```
shalins, line 5: Why is color the only thing in props? Do we plan on adding more? Additionally, to me it seems that `BoxProps` and `CylinderProps` would contain the same fields, no? If so they should be combined into a single interface and placed in a separate file. Then import it here.

src/Shapes/Box.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { BoxProps, useBox } from '@react-three/cannon';
+
+interface CustomBox extends BoxProps {
```
shalins, line 4: I'm still not sure what the point of `Props` is, which makes it hard for me to extend it in the future. Please add a comment explaining what the intention and use of this interface is.

src/Shapes/Box.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * @returns
+ */
+export default function Box(props: CustomBox) {
+  const [collisionRef] = useBox(() => ({
+    args: props.args,
+    mass: props.mass,
+    position: props.position,
+    type: props.type,
+    onCollide: props.onCollide,
+    ...props,
```
shalins, line 21: I get why you added `props` at the end is, however generally good convention is that in a list of arguments where one of them is a closure, you pass the closure as the final argument, that way the caller has clean and obvious syntax. I would find a way to move `onCollide` as the final argument.

src/Shapes/Box.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface CustomBox extends BoxProps {
+  color: string;
+}
+/**
+ * Box component renders a box with collision detection.
+ * This component accepts props that determine its size, position, type, and mass,
+ * and a callback to be called upon a collision.
+ * @param props args: Array<number>, mass: number, position: Triplet, type: string, onCollide: Function
+ * @returns
```
shalins, line 12: Remove this if the function returns nothing. Else, fill it out with a description.

src/Shapes/Cylinder.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface CustomCylinder extends CylinderProps {
+  color: string;
+}
+/**
+ * Cylinder component renders a cylinder with collision detection.
+ * This component accepts props that determine its size, position, type, and mass,
+ * and a callback to be called upon a collision.
+ * @param props args: Array<number>, mass: number, position: Triplet, type: string, onCollide: Function
+ * @returns
```
shalins, line 12: Again, there is no description here even though this function returns a cylinder mesh.

src/Shapes/Cylinder.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { CylinderProps, useCylinder } from '@react-three/cannon';
+
+interface CustomCylinder extends CylinderProps {
+  color: string;
+}
+/**
+ * Cylinder component renders a cylinder with collision detection.
+ * This component accepts props that determine its size, position, type, and mass,
+ * and a callback to be called upon a collision.
+ * @param props args: Array<number>, mass: number, position: Triplet, type: string, onCollide: Function
```
shalins, line 11:
```suggestion
 * @param props: Array<number>, mass: number, position: Triplet, type: string, onCollide: Function
```

src/Shapes/Cylinder.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { CylinderProps, useCylinder } from '@react-three/cannon';
+
+interface CustomCylinder extends CylinderProps {
+  color: string;
+}
+/**
+ * Cylinder component renders a cylinder with collision detection.
+ * This component accepts props that determine its size, position, type, and mass,
```
shalins, line 9:
```suggestion
 * Accepts props that determine its size, position, type, and mass,
```

src/Shapes/Cylinder.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { CylinderProps, useCylinder } from '@react-three/cannon';
+
+interface CustomCylinder extends CylinderProps {
+  color: string;
+}
+/**
+ * Cylinder component renders a cylinder with collision detection.
```
shalins, line 8:
```suggestion
 * Renders a cylinder component with collision detection.
```

src/Shapes/GLTFCollision.tsx
```diff
@@ -0,0 +1,49 @@
...
+  bboxScaleFactor: number
+  modelPath: string
+}
+
+/**
+ * GLTFCollisionModel component renders a GLTF with a bounding box for collision detection.
+ * This component accepts props that determine its size, position, type, 
+ * a callback to be called upon a collision, path to GLTF to load, and scaling factors for the model and bbox.
+ * @param props args: Array<number>, mass: number, position: Triplet, type: string, onCollide: Function,
+ * modelScaleFactor: number, bboxScaleFactor: number, modelPath
```
shalins, line 20: Can we make this a bit more descriptive if possible? Additionally, I wouldn't add this specific of a description. For example, adding `Array<number>, mass: number, position: Triplet, type: string, onCollide: Function` makes this brittle since if you change the `CustomGLTF` object to now include different fields, then all of your comments would be out of date and wrong. Instead, mention that `props` is of type `CustomGLTF`

src/Shapes/GLTFCollision.tsx
```diff
@@ -0,0 +1,49 @@
...
+ * @returns
+ */
+export default function GLTFCollisionModel(props: CustomGLTF) {
+  const { scene } = useLoader(
+    GLTFLoader,
+    props.modelPath,
+  );
+  // Must use Skeleton utils to support copies of skinned meshes https://github.com/mrdoob/three.js/issues/11573
+  let copiedScene = useMemo(() => SkeletonUtils.clone(scene), [scene]);
+  copiedScene.scale.multiplyScalar(props.modelScaleFactor);
```
shalins, line 30: Nice use of constants!

src/Shapes/Sphere.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { SphereProps, useSphere } from '@react-three/cannon';
+
+interface CustomSphere extends SphereProps{
```
shalins, line 4:
```suggestion
interface CustomSphere extends SphereProps {
```

These things should be fixed by using a linter when coding.

shalins (changes requested):

src/Player/playerMovementEmitter.ts
kanyesthaker: yeah this is strange, what this means is that our linter configs aren't shared -- @samjgorman is your `eslintrc.json` the same as the one in `main`? did you follow the same steps as the ones listed in the README?

One solution (since there may be independent VSCode config differences that are affecting eslint behavior) is to add a pre-commit hook so that we run a single shared linter on every `git push` -- along with a `--skip-lint` flag if you want to disable it; I'll look into adding this in the near future, but for now if we can't debug this easily we shouldn't block on this, a little ugliness at this stage is fine so long as there's a high prio plan to fix it (i.e. if we can't figure out a solution right now w/ comparing `eslintrc`s, we shouldn't stop committing code until we do).

kanyesthaker: At work I generally just ignore linting throughout my dev workflow, then run the linter at the very end before I ask for a PR review and merge into main

src/Scene/Scene.tsx
kanyesthaker: +💯 we should configure the linter to do this

src/Scene/Scene.tsx
kanyesthaker: afaik this is temporary since we didn't have classes for lighting or physics yet (this is tech debt), and wanted to get the base scene in `main` to unblock dev. Don't need to fix this in this PR (since we have tickets to track these in the future) but good catch and smart callout @shalins

src/Scene/Scene.tsx
kanyesthaker: good idea; in fact, I'd probably go a step further and handle the URL parsing as an arg to `getS3Path`, i.e. you should be able to do something like
`const tropicalPlant = getS3Path('plant', 'gltf');`
`const skyboxTexture = getS3Path('skybox', 'texture');`
etc., and handle the actual path construction as a utility

src/Scene/Scene.tsx
kanyesthaker: yeah file naming is annoying, since adding descriptors as part of a filepath will also get messy (in general because two lines of code accessing a file may not use it for the same reason). But I do agree with having some kind of constant file dictionary that maps filenames to proper descriptors!

src/Scene/Scene.tsx
samjgorman: Can you take the lead on this @kanyesthaker?  Agree with the above.  This can be in a future cleanup PR, or you can add it within the existing PR.

src/Scene/Scene.tsx
```diff
@@ -18,17 +19,106 @@ export default function Scene() {
   return (
     <div style={{ height: window.innerHeight }}>
       <Canvas>
```
kanyesthaker, line 21: We should consider how we're thinking about this `return` block on `Scene.tsx` -- as we can see as we add more objects it's quickly expanding to a point where we're losing some readability. I'd probably prefer something like creating a `List` of objects -- i.e. a list of `Object` components, a list of `Lighting` components, etc, and then simply returning something analogous to

```
return(<div...><Canvas>{LightingComponents}{ObjectComponents}...</Canvas></div>)
```

src/Scene/Scene.tsx
samjgorman: I can do this!

src/Scene/Scene.tsx
samjgorman: This is part of the BoxProps I extended in the CustomBox interface.   The `UseBox` hook expects certain properties, such as `args`, `mass`, `position` and` type`.  The `ref` that UseBox returns allows us to "hook" into Cannon.js's physics engine.  So, I think we'll need to keep args in the actual component, but can abstract the name to something more idiomatic when the user provides a prop by creating a property in `CustomBox`.

src/Scene/Scene.tsx
```diff
@@ -18,17 +19,106 @@ export default function Scene() {
...
+          type={'Static'}
+          onCollide={() => console.log('I also collided!')}
+          modelScaleFactor={0.3}
+          bboxScaleFactor={0.}
+          modelPath={getS3Path('models/mushroom_tree/scene.gltf')}
+          /> 
+
+
+        
+       <GLTFCollisionModel
```
kanyesthaker, line 97: Appreciate the reusability here -- nice use of component-ization. I'm rethinking our earlier discussion about creating the component class based on collision -- it feels more intuitive to me that collision should be a property of an object, since not every object will have collision (i.e. i think a fern on the ground shouldn't have collision).  Would it make more sense to have modify the existing `SceneObject` component to accept collision as a boolean prop? I additionally am wary of having a long list of objects in the Scene that all start with `<GLTFCollisionModel>` -- i'm envisioning something like this:

```
var plants = () => {
    return (
        <Object name={plantName1} collision={false} position={[1,2,3]}.../>
        <Object name={plantName2} collision={false} position={[1,2,3]}.../>
        <Object name={plantName3} collision={false} position={[1,2,3]}.../>
}
```
and then the `Scene()` function simply returns
```
return (
    <div...>
        <Canvas>{plants}</Canvas>
    </div>
)
```

src/Scene/Scene.tsx
samjgorman: We'll call this property `dimensions` to correspond to the x,y,z dimensions this array of number provides?

src/Shapes/GLTFCollision.tsx
samjgorman: Thanks!

src/Scene/Scene.tsx
samjgorman: Feel free to correct me here, but I believe doing so would require us to place the Box, Cylinder, and Sphere components into one shapes.tsx file or use a more complex workaround.   This is not optimal, since it's more conventional to have each single React component in a single file.  At the very least, I'll group these imports next to each other in the code. 

See more:
https://stackoverflow.com/questions/29722270/is-it-possible-to-import-modules-from-all-files-in-a-directory-using-a-wildcard

src/Shapes/Sphere.tsx
samjgorman: Great!

src/Scene/SceneObject.tsx
kanyesthaker: agree; also generally wary of the pattern
`if not condition: return` -- much prefer `if condition: do_something // else: do_something_else`

src/Scene/Scene.tsx
samjgorman: Agree!  Let's keep `Scene.tsx` as high level as possible.  This should be addressed in the next PR that places objects, lighting or shadows into the world by creating child components that act as containers for multiple relevant components.  Whether they are rendered in a list is likely implementation-specific, so I'll leave that to future work

src/Shapes/Box.tsx
kanyesthaker: is there a docstring standard for es6? should just follow that if there is

src/Scene/Scene.tsx
samjgorman: Makes sense!

src/Scene/SceneObject.tsx
samjgorman: Do we still need this file?  I'm fine keeping it in the project repo if you think it may hold reference value, but won't make any changes to it.  However, let's keep this convention in mind moving forward and bias towards `if condition: do_something // else: do_something_else`

src/Scene/SceneObject.tsx
samjgorman: Makes sense, I'll address magic numbers more generally in another suggestion of yours since this file is less relevant to the project now

src/Scene/SceneObject.tsx
shalins: Actually @kanyesthaker in this case I do prefer the `if NOT condition` first -- reason being is that if there's a case where we're returning early the reader should know about that before the reading the rest of the logic.

src/Shapes/Box.tsx
kanyesthaker: I believe that `...props` has to be at the end of the param list -- `...` unpacks the list, but it has the special property that when a `...` parameter (a **spread** parameter) occurs at the end of the parameter list, it functions as a **rest** parameter (it gathers all remaining parameters into a list). More information [here](https://javascript.info/rest-parameters-spread) -- if we don't need `rest` functionality here we can reorganize the parameters.

src/Scene/Scene.tsx
shalins: Wildcard is not great because is pollutes the global namespace. Instead, what you can do is put them in one file, then expose them one at a time using named imports. Another alternative is to use separate files and then use a file whose entire code is just exporting the functions to the public. It's up to you how to implement the specifics, however I do recommend grouping together for the user of those functions.

src/Shapes/Cylinder.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { CylinderProps, useCylinder } from '@react-three/cannon';
+
+interface CustomCylinder extends CylinderProps {
```
kanyesthaker, line 4: as before, some more color on what exactly `CylinderProps` does would be helpful!

src/Scene/Scene.tsx
shalins: I agree! At the very least let's practice creating locally namespaced constants when we want to prototype! Keeping the tech debt to the minimum is great when the time cost is pretty minimum.

src/Shapes/GLTFCollision.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useMemo } from 'react';
+import { GLTFLoader } from 'three/examples/jsm/loaders/GLTFLoader';
+import { useLoader } from '@react-three/fiber';
+import { Box3 } from 'three';
```
kanyesthaker, line 4: echoing Sam's import comments from above! just commenting for visibility

src/Shapes/GLTFCollision.tsx
```diff
@@ -0,0 +1,49 @@
...
+export default function GLTFCollisionModel(props: CustomGLTF) {
+  const { scene } = useLoader(
+    GLTFLoader,
+    props.modelPath,
+  );
+  // Must use Skeleton utils to support copies of skinned meshes https://github.com/mrdoob/three.js/issues/11573
+  let copiedScene = useMemo(() => SkeletonUtils.clone(scene), [scene]);
+  copiedScene.scale.multiplyScalar(props.modelScaleFactor);
+
+  let bbox = useMemo(() => new Box3().setFromObject(copiedScene), [copiedScene]);
```
kanyesthaker, line 32: lots happening here -- nice work with brevity. Would love a few comments here to make it easier to track the logic!

src/Scene/Scene.tsx
samjgorman: So... we can modify this to accept a collisions bool with modifications like this. I think this is reasonable for now, but you can see how this sort of conditional rendering can quickly become out of hand if there exist multiple properties that modify what is rendered in the component. So, we can rename this to GLTFModel, and remove the prior GLTFModel component we had

```typescript
export default function GLTFCollisionModel(props: CustomGLTF) {
  const { scene } = useLoader(
    GLTFLoader,
    props.modelPath,
  );
  // Must use Skeleton utils to support copies of skinned meshes https://github.com/mrdoob/three.js/issues/11573
  let copiedScene = useMemo(() => SkeletonUtils.clone(scene), [scene]);
  copiedScene.scale.multiplyScalar(props.modelScaleFactor);
  let collisionRef;
  if (props.collision){
    let bbox = useMemo(() => new Box3().setFromObject(copiedScene), [copiedScene]);
    const bboxSize = bbox.getSize(new Vector3());
    const scaledBbox = bboxSize.multiplyScalar(props.bboxScaleFactor).toArray();
    collisionRef= useBox(() => ({
      args: scaledBbox, //Must accept array and not vector3
      mass: props.mass,
      position: props.position,
      type: props.type,
      onCollide: props.onCollide,
    }));
  }
 

  return (
    <> 
    { props.collision ? <primitive ref={collisionRef} position={props.position} object={copiedScene} dispose={null} /> :   <primitive position={props.position} object={copiedScene} dispose={null} />}

    </>
  );
}
```

src/Shapes/Cylinder.tsx
samjgorman: Why? Curious if theres a convention i dont know

src/Scene/Scene.tsx
samjgorman: As an update: I've added a collision prop to each shape. So, each shape can either use collisions or not! Check out the code when you review again @kanyesthaker

src/Scene/Scene.tsx
samjgorman: This is another required prop for our physics engine. We can create an abstraction for this.  Have any ideas for alternatives to string?

More context on what these types mean:

`dynamic`: A dynamic body is fully simulated. Can be moved manually by the user, but normally they move according to forces. A dynamic body can collide with all body types. A dynamic body always has finite, non-zero mass.

`static`: A static body does not move during simulation and behaves as if it has infinite mass. Static bodies can be moved manually by setting the position of the body. The velocity of a static body is always zero. Static bodies do not collide with other static or kinematic bodies.

`kinematic`:  kinematic body moves under simulation according to its velocity. They do not respond to forces. They can be moved manually, but normally a kinematic body is moved by setting its velocity. A kinematic body behaves as if it has infinite mass. Kinematic bodies do not collide with other static or kinematic bodies.

See more: 
https://github.com/pmndrs/use-cannon/tree/master/packages/react-three-cannon

src/Shapes/GLTFCollision.tsx
samjgorman: done

src/Scene/Scene.tsx
samjgorman: Can we move this to a linter PR over the next few days?  Would prefer to decouple logic of collisions from linter, and this sounds like more of a team effort while collisions is mostly an area I've been leading.

samjgorman: I don't think we should be moving these to one file as this sacrifices readability & decomp of components, but I'll look into the second idea you mentioned + other potential approaches

samjgorman: Implemented with a variation of the second idea!

Can do this in an `index.ts i`n `shapes`

```typescript
import Box from './Box';
import Cylinder from './Cylinder';
import Sphere from './Sphere';
import GLTFModel from './GLTFModel';


export { Box, Cylinder, Sphere, GLTFModel };
```

Nice idea to simplify

samjgorman:
Ok, I've addressed most of the comments listed above @shalins @kanyesthaker.  Notable exceptions are linter issues, which we're currently discussing & will likely introduce a new PR after this that directly addresses configuring the linter. To make your lives easier, here's high-level overview of changes

1. Each component in shapes library now accepts a collision boolean prop that determines whether collisions are active.  To this end, GLTFCollisionModeland GLTFModel were merged 
2. Shapes library can be imported in one line 
3. `worldConstants.ts` introduced, can become a store for constants related to world-building & physics. 
4. More clarity in docs & naming when possible.  There's more work that can be done to abstract some of the more confusing parts of the physics library required props, but trying to strike a balance between speed & clarity.

src/Scene/Scene.tsx
samjgorman: Appreciate your attention to detail here.

Modern browsers can get HEX and RGB from named colors programatically without us needing to construct a UI file! As a toy example, I passed grey and gray as props to two different objects and the browser was able to resolve successfully. 

Having something like a UI file of allowed colors may make sense down the line, especially as we look to assert more control over the visual styling of the application, but I'm cautious about introducing this at this stage. 

image

src/Scene/Scene.tsx
samjgorman: image

More semantic names for model objects!

src/Scene/Scene.tsx
kanyesthaker: yeah that's fine -- as a general rule of thumb, at work the policy is "if you see tech debt, clean it up" even if it's not part of the main feature you're shipping -- however, since there's so much that's unclear about the actual solution to this linting issue we can work on it separately

src/Scene/Scene.tsx
kanyesthaker: i can do this

src/Scene/Scene.tsx
shalins: Yeah I'll leave the implementation details up to you!

src/Scene/Scene.tsx
shalins: Nice! Thanks for verifying

src/Scene/Scene.tsx
shalins: Yeah generally things like this are a good place to use enum. It's up to you whether or not you want to create an abstraction for this. My take is that if it's something that's used often enough, it increase the likelihood that someone would mess up the string (typo, etc.). The other case is if you're creating an interface that other devs would be using, since enums are cleaner to use.

src/Scene/Scene.tsx
shalins: Thanks!

src/Scene/SceneObject.tsx
shalins: As an update, good practice for returning early is something like this:
```

# Divides `a` by `b`
def divide(a, b):
    if b == 0:
         # Return early if b is zero.
        return
    return a / b
```

instead of:
```
# Divides `a` by `b`
def divide(a, b):
    if b == 0:
        return
    if b != 0:
        return a / b
```

src/Shapes/Box.tsx
shalins: I found this:
```
/**
 * @name hash
 * @description Random hashing algorithm I found on Stack Overflow.
 * @param {string} str 
 * @param {boolean} asString 
 * @param {*} seed 
 * 
 * @returns {string} hash
 */
function hash(str, asString, seed) {
```

src/Shapes/Cylinder.tsx
shalins: Not a convention, though it's good practice to make sure your comments are a) as short as possible without b) losing clarity. In this case the "This component" is not needed. When commenting on a function, no need to say "This function does X". Can just say "Does X." Same thing with variables.

src/Scene/Scene.tsx
samjgorman: Great, I introduced a dimensions prop that is more descriptive than args for this case!

src/Scene/Scene.tsx
samjgorman: okay! thanks for clarifying here :)

shalins:
I've finished responding to some of your comments @samjgorman. Let me know once you're done making the changes and resolve all the existing comments, then after a quick second pass you should be good to merge! Sorry this was a long review.

src/Scene/Scene.tsx
samjgorman: Thanks for explaining! After learning more I agree that it makes sense to make this an `enum` (hadn't used them a lot before)! I'll get this done

samjgorman: As an update, this has already been typed for us as part of the cannon physics module, which means I don't need to construct a new enum.  In other words, the only acceptable string names are what are listed above, and will display an error if you attempt to write a different string name.  So, this is essentially the enum we've discussed. However, for cases where it has not yet been typed for us, it's a great idea to do what you've discussed! thanks for the tip

src/Scene/Scene.tsx
samjgorman: Marking as resolved after introducing` worldConstants.ts` for this purpose

src/Shapes/Cylinder.tsx
samjgorman: Added comments for each type!

src/Scene/SceneObject.tsx
kanyesthaker: Just to avoid confusion, in the above example the best format is
```
def divide(a, b):
    try:
        return a / b
    except ZeroDivisionError:
        print('Cannot devide by zero.')
```
(you don't want to hide something as major as a divide-by-zero!)
Nitpicking aside, after a second thought you're right @shalins, it does make more sense to add the early exit clause before the remaining logic (StackOverflow also backs you up https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement)

src/Shapes/GLTFModel.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+interface CustomGLTF extends BoxProps {
+  modelScaleFactor: number;
+  bboxScaleFactor: number;
+  modelPath: string;
+  collision: boolean;
+}
+
+/**
+ * GLTFCollisionModel component renders a GLTF with a bounding box for collision detection.
```
kanyesthaker, line 17: this docstring needs to be updated

src/allTypes.ts
```diff
@@ -8,4 +8,11 @@ export type PlayerVelocity = {
   x: number;
   y: number;
   z: number;
-};
\ No newline at end of file
+};
+
+export enum Directions {
```
kanyesthaker, line 13: nice

src/Shapes/GLTFModel.tsx
samjgorman: done!

src/Player/playerMovementEmitter.ts
samjgorman: A note before I resolve this: I reintroduced prettier for code formatting to my workflow, which should fix most lout of sync linting issues for future PRs.  The no comma dangle rule was conflicting across ESLint and Prettier, so I added a line in eslint's config file that turns off enforcing comma-dangle rule such that conflicts between ESlint and prettier are automatically resolved. We'll address more robust options for enforcing linting consistency in a future PR!

```typescript
        "comma-dangle": 0,
        "@typescript-eslint/comma-dangle": 0
```

src/Shapes/Box.tsx
samjgorman: Fun tip: can generate a docstring skeleton for your function in VS code by writing /** then pressing enter above your function! This is what I used to write our docstrings.

src/Shapes/Cylinder.tsx
samjgorman: addressed

src/allTypes.ts
samjgorman: This Directions enum is actually not used in the production application yet, so I actually removed this from types. That said, we now have a helpful template for introducing enums that we can use in future work

kanyesthaker (approved):
lgtm

shalins (approved):

PR #13, opened 2022-04-17
Sam/world boundaries

samjgorman:
### Description of Changes
Introduce world boundaries with addition of boundaries.tsx and modification of floor.tsx. 
Refactor shapes library for clarity, introduce new props to make shapes translucent.
Introduced a changelog for us to maintain over time.

image

### Link to Notion ticket

### Testing Procedure
Tested each prop in boundaries & shapes lib locally in combination with other props to ensure expected behavior. Modified world constants for height and width to ensure that boundaries resize correctly.


samjgorman:
@kanyesthaker  @shalins take a look!

changelog.md
```diff
@@ -0,0 +1,11 @@
+# Changelog
```
kanyesthaker, line 1: are we doing a manual changelog? github autogenerates one which we can use, esp bc we're not really tracking releases (each PR shouldn't be its own release)

changelog.md
```diff
@@ -0,0 +1,11 @@
...
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [1.0.0] - 2022-04-16
```
kanyesthaker, line 7: should be v0.1.0 i believe

src/Scene/Boundaries.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import { PlaneProps } from '../allTypes';
+import Box from '../Shapes/Box';
```
kanyesthaker, line 3: hmmm, I appreciate the reuse of code here to avoid re-implementation, but using 4 boxes as the world boundaries feels a little clunky

src/Scene/Boundaries.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Boundaries functional component renders four boundaries of a plane given
```
kanyesthaker, line 6: Let's make this a little more functionally descriptive -- explain what the world boundary is vs. exactly how it's implemented

kanyesthaker: for example:

```
Creates a world boundary (4 impassible planes) that define the edges of the world...
```

src/Scene/Boundaries.tsx
```diff
@@ -0,0 +1,63 @@
...
+ * Boundaries are fullytranslucent unless debug prop is set to true.
+ * @param props: PlaneProps
+ * @returns
+ */
+function Boundaries(props: PlaneProps) {
+  let boundaryColor;
+  let boundaryHeight;
+
+  if (!props.boundaryHeight) {
+    boundaryHeight = 50;
```
kanyesthaker, line 17: is there any reason for this to not be a constant? why would we need this to be a prop?

src/Scene/Boundaries.tsx
```diff
@@ -0,0 +1,63 @@
...
+      color: 'yellow',
+    };
+  } else {
+    boundaryColor = {
+      transparent: true,
+    };
+  }
+  return (
+    <>
+      <Box
```
kanyesthaker, line 33: Instead of using 4 boxes, why don't we either
a) use 4 planes to define the world boundary
b) use 1 _really big_ box (collision should work whether you're on the outside or the inside of a shape) (the skybox is effectively option (b))

changelog.md
samjgorman: Modified!

samjgorman:
Ok @kanyesthaker @shalins , should be ready for re-review. Here's a summary of changes:

1. Introduced a `Plane` component to the shapes library.
2. Refactored `Boundary.tsx` to use 4 impassable positioned & rotated `Plane` components.  This is better than using the `Box` shape because planes can extend infinitely in 2D space to act as impassable boundaries, and removes the need to specify a height. 
3. Refactored `Floor.tsx` to use this `Plane` component.
4. Some QoL updates such as introducing default values for the `type` prop to cut down on the number of required props necessary to render components in the shapes library.

src/Scene/Boundaries.tsx
samjgorman: Okay, I went with option A after doing more research.  I haven't seen a lot of successful examples of what we're attempting to do online, so happy to look through any resources you come across, but I agree that using 4 planes rather than boxes is more robust & eliminates the need to specify a `boundaryHeight`.

src/Scene/Boundaries.tsx
samjgorman: This is no longer necessary! Happy to reduce complexity here.

src/Scene/Boundaries.tsx
samjgorman: Great idea! Modified.

src/Scene/Boundaries.tsx
samjgorman: Agreed! Resolving this after implementing a `plane` based approach

changelog.md
samjgorman: I would _like_ to do a manual changelog to get experience writing a changelog & provide a more concrete ledger of our progress over time.  Perhaps we should remove / modify the concept of releases in the changelog and simplify keep a txt file that we update each week with the major changes we shipped?

src/Player/Player.tsx
```diff
@@ -37,15 +37,15 @@ export default function Player(props: SphereProps) {
     setPlayerRef.velocity.subscribe((playerVelocity) => {
       currentVelocityVector.current.x = round(
         playerVelocity[0],
-        ROUNDING_PRECISION,
+        ROUNDING_PRECISION
```
samjgorman, line 40: Note: Looks like a linting issue here. I think this will be resolved in future PRs that use your prettier config, the updated eslintc.json, and a pre-commit hook.

src/Scene/Scene.tsx
```diff
@@ -106,15 +109,13 @@ export default function Scene() {
...
             position={[50, 1, 65]}
-            type={'Static'}
             modelScaleFactor={0.3}
             bboxScaleFactor={0.5}
             modelPath={getS3Path('models/mushroomTree/scene.gltf')}
             collision={true}
           />
 
-          <Floor />
-          <gridHelper args={[100, 100, 'black', 'grey']} />
```
samjgorman, line 117: Note: Will need to go back and make sure this code uses the constants you've set up in your PR once we've gone and merged

src/Shapes/Plane.tsx
```diff
@@ -0,0 +1,64 @@
...
+  rotation?: [number, number, number];
+  collision?: boolean;
+}
+
+/**
+ * Plane component renders a plane with collision detection.
+ * @param props customPlane
+ * @returns
+ */
+export default function Plane({ type = 'Static', ...props }: CustomPlane) {
```
samjgorman, line 22: One more note: I also didn't modify the structure of the components in the shapes library in this PR, but looking forward to explore approaches to simplify & extend these components in a future PR

src/Scene/Scene.tsx
samjgorman: Constants now used after merging the constants PR and resolving merge conflicts

src/Shapes/Box.tsx
```diff
@@ -17,34 +18,45 @@ interface CustomBox extends BoxProps {
...
-        <mesh position={props.position}>
-          <boxBufferGeometry args={props.dimensions} />
-          <meshPhongMaterial color={props.color} />
-        </mesh>
-      )}
+      <mesh {...meshProps}>
+        <boxBufferGeometry args={props.dimensions} />
+        <meshPhongMaterial
+          color={props.color}
+          transparent={transparent}
```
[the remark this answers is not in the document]
samjgorman: Great idea! Similar theme here to video UI PR, great lessons going forward

src/Shapes/Box.tsx
```diff
@@ -28,23 +38,25 @@ export default function Box(props: CustomBox) {
...
-        </mesh>
-      )}
-
-      {!props.collision && (
-        <mesh position={props.position}>
-          <boxBufferGeometry args={props.dimensions} />
-          <meshPhongMaterial color={props.color} />
-        </mesh>
-      )}
+      <mesh {...meshProps}>
```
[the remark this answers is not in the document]
samjgorman: I've refactored how meshProps is constructed and hope that this provides more clarity! Now, meshProps looks like this (implemented across the shapes lib)

```typescript
const meshProps = {
    ref: props.collision ? collisionRef : undefined,
    position: props.position,
  };
```

src/Scene/Scene.tsx
```diff
@@ -106,15 +109,13 @@ export default function Scene() {
...
+          <Floor lengthX={PLANE_LENGTH_X} widthZ={PLANE_WIDTH_Z} />
```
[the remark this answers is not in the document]
samjgorman: Thanks for the feedback!  We now have a `WorldBox` component that functions as a simple wrapper over the `Floor` & `Boundary` components. 

I agree that they should not exist without each other, but would like to keep each component as simple or "pure" as possible.  For example, the `Floor` component may grow in complexity as we introduce new textures or varied terrain, which may become less readable if combined with the `Boundaries` component. 

And in this approach, we still gain the benefits of passing shared props such as `debug` 

Would love your opinion on...
1. Does `WorldBox` make sense for the name of this component?  I see this component as a box extending across the horizontal plane & each four edges.

changelog.md
samjgorman: I removed this txt file in favor of a Notion page I'll spin up!

samjgorman:
Ready for rereview @kanyesthaker!

kanyesthaker (approved):
small nit, otw lgtm

src/Scene/WorldBox.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Renders a WorldBox: made from a Boundaries component & Floor component.
+ * Boundaries renders 4 impassible Planes that define the edges of the world,
+ * and Floor renders a horizontal, impassable Plane representing the
+ * lowest reachable position in the Y direction.
+ * @param props
+ * @returns WorldBox component
+ */
+const WorldBox = (props: WorldBoxProps) => {
```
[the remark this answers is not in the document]
samjgorman: I think world boundaries is more accurate-- updated! While I was there I updated how we handled props in our shapes library and we now destructure props before using them, which is better style :)

PR #14, opened 2022-04-19
Sam/video UI

samjgorman:
### Description of Changes
Video UI with minimal styling for local audio / video with ability to toggle audio / video states.

Uses browser's navigator.mediaDevices.getUserMedia API to obtain video and or audio streams upon component mount & upon changes to video or audio state.

Screen Shot 2022-04-19 at 00 00 01

### Link to Notion ticket
https://www.notion.so/irisapp/Video-player-UI-dbb7addf6d9a4af89df402ede590b472

### Testing Procedure
Tested each toggle combination in browser for expected behavior, such as correctly toggling on / off in the browser UI & user's webcam. Tested mostly in Chrome and somewhat in Safari. More work needs to be done to ensure this works across browsers.


samjgorman:
Ready for review! @shalins @kanyesthaker

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+    return newPermissions;
+  }
+
+  useEffect(() => {
+    async function getMedia() {
+      try {
+        let permissions = configurePermissions();
+        startMedia(permissions);
+      } catch (err) {
+        // TODO: handle common error status described here
```
shalins, line 141:
```suggestion
        // TODO(Sam): handle common error status described here
```

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+      newPermissions = {
+        video: true,
+      };
+    } else {
+      newPermissions = {
+        video: true,
+        audio: true,
+      };
+    }
+    return newPermissions;
```
shalins, line 132: Can we simplify?
```suggestion
    let newPermissions = {
            video: videoOn,
            audio: audioOn,
    };
```

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+        width={props.width}
+        autoPlay
+      ></Video>
+
+      <IconContainer>
+        {audioOn ? (
+          <Microphone onClick={handleAudio} icon={faMicrophone} />
+        ) : (
+          <MicrophoneSlash onClick={handleAudio} icon={faMicrophoneSlash} />
+        )}
```
shalins, line 197:
```suggestion
        {(
          <Microphone onClick={handleAudio} icon={audioOn ? faMicrophone : faMicrophoneSlash} />
        )}
```

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+        {videoOn ? (
+          <VideoIcon onClick={handleVideo} icon={faVideo} />
+        ) : (
+          <VideoIconSlash onClick={handleVideo} icon={faVideoSlash} />
+        )}{' '}
```
shalins, line 202: Please keep ternary operators as short as possible when using them, otherwise they can obfuscate the code. Try to keep it < 1 line. If it's > 1 line, use normal if statements instead.

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+ * in useEffect hooks invoked upon component mount & upon change in audioOn or videoOn state.
+ * The srcObject of the video Ref is then set to the stream returned from the navigator browser API.
+ *
+ * On a state change update to end video or audio, endMedia is invoked with appropriate type specified.
+ * This function then stops each appropriate live track to end the stream & turn off the user's webcam light.
+ *
+ * On a state change update to start video or audio, endMedia is invoked again to clear any current active streams,
+ * then startMedia is invoked to receive a stream from the navigator browser API and reset videoRef's srcObject to this stream.
+ * @param props VideoProps
+ * @returns
```
shalins, line 78: Please fix your doc comments. Add return value if there is one else remove the `@returns`

samjgorman: Thanks for pointing this out! I can make this more functionally descriptive.

I did more research on what was conventional and seems like when a function is `void`, removing the `@returns` is an acceptable choice.  I'll update. 

https://stackoverflow.com/questions/4759175/how-to-return-void-in-jsdoc

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+import styled from 'styled-components';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import {
+  faMicrophoneSlash,
+  faMicrophone,
+  faVideo,
+  faVideoSlash,
+} from '@fortawesome/free-solid-svg-icons';
+
+const VideoContainer = styled.div`
```
shalins, line 11: What's the difference between `VideoContainer` and `Video`? Could we name it more clearly?

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+  position: absolute;
+  z-index: 999;
+  display: flex;
+  flex-direction: column;
+`;
+
+const Video = styled.video`
```
shalins, line 18: Can we move these to a diff file? Video.tsx is already 200+ lines of code.

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+        stream.getTracks().forEach(function (track) {
+          if (trackKind === 'all') {
+            if (track.readyState == 'live') {
+              track.stop();
+            }
+          } else {
+            if (track.readyState == 'live' && track.kind === trackKind) {
+              track.stop();
+            }
+          }
```
shalins, line 99:
```suggestion
          if (trackKind === 'all' && track.readyState == 'live') {
              track.stop();
          } else if (track.readyState == 'live' && track.kind === trackKind) {
              track.stop();
          }
```

You don't need to write `if` statements nested like this, you can simplify. That being said, the second issue with the code is lack of clarity in intention. 

A general pointer -- wherever you find yourself writing complex logic, take a pause and inquire whether there is an easier or more clear way to get the job done. There often is. Once you get it done, document like hell. 

Code is not just for computers to read. It's also equally, if not more so, for humans to read! Humans are the ones that use it, modify it, refactor it, extend it. If someone with no prior context can come in and understand your code, you've done a good job (assuming it actually works as intended).
If someone with know prior knowledge

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+  border-radius: 32px;
+`;
+
+const MicrophoneSlash = styled(FontAwesomeIcon)`
+  color: #ed4650;
+  margin: 2%;
+`;
+
+const Microphone = styled(FontAwesomeIcon)`
+  color: black;
```
shalins, line 38: Please keep the format consistent. Either use hex or use strings, but ideally don't switch between both.

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+  margin: 2%;
+`;
+
+const VideoIconSlash = styled(FontAwesomeIcon)`
+  color: #ed4650;
+  margin: 2%;
+`;
+
+type VideoProps = {
```
shalins, line 52: Don't use `typealias` here, use `interface` instead. IMO, type aliases should be used for creating simpler or clearer names for longer types. Interfaces should be used when defining a structure or skeleton for a type. Since this is the latter, I'd use an interface instead.

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+  height?: number;
+  width?: number;
+};
+
+enum VideoAudioOptions {
+  audio = 'audio',
+  video = 'video',
+  all = 'all',
+}
```
shalins, line 61:
```suggestion
enum AVOptions {
  AUDIO,
  VIDEO,
  BOTH,
}
```

Also you can just index into the enum to get a string value, no need to assign it (`AVOptions[AVOptions.AUDIO]`). Also for enums you can use a switch statement or compare it by the actual variant `trackKind === AVOptions.BOTH`. Don't compare the string value itself. That defeats the whole point of using enums.

shalins (dismissed):
Overall, nice work! 

My feedback is that I'd love to see more intention in the code. Right now, there are still some basic things I have to comment on (complicated if-statement logic, incomplete doc comments, etc.) and it doesn't feel like there's a strong reason for doing things the way you're doing them. 

I'd recommend thinking through each line of code you're adding and really consider if it's necessary. Could you simplify it? Think through the design of your small changes, as well as the system as a whole. For example, how should you scope different elements? What should their access level be? What are clear variable and function names? How to split up the files? etc. It's much like writing a great essay! Go through a first, second, third pass on your code and make sure you're getting all the basics down before pushing a PR.

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * VideoPlayer component renders configurable video & audio media from a user's device.
```
shalins, line 64:
```suggestion
 * `VideoPlayer` renders configurable video & audio media from a user's device.
```

src/App.tsx
```diff
@@ -1,9 +1,13 @@
...
 import Scene from './Scene/Scene';
+import VideoPlayer from './Video/video';
 
 function App() {
   return (
-    <Scene/>
+    <>
+      <VideoPlayer height={115} width={150} />
+      <Scene />
+    </>
```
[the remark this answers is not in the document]
kanyesthaker: agree, would like a `<Display/>` component that contains things like `<VideoPlayer/>`, `<Controls/>`, `<Chat/>` etc. (then controls is further broken down into various buttons, etc.)

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+ * Renders UI to toggle audio & video on & off.
+ *
+ * The component uses the useRef hook to obtain a reference to a <video/> component.
```
shalins, line 67:
```suggestion
 * Uses the `useRef` hook to obtain a reference to a `<video/>` component.
```

src/Video/video.tsx
kanyesthaker: +1

src/Video/video.tsx
shalins: See: https://initialcommit.com/blog/code-documentation

Some notable quotes:

- "When writing documentation it is helpful to ask yourself, 'Will someone who has no knowledge of the subject understand exactly what I'm saying?'"
- "The best documentation always has the end-user top of mind. This statement holds true in all cases, as there is no point in writing documentation if no one even understands what you are writing. You're better off not writing at all!"
- "By adding that one extra line, we now have some context into why they authored the code in the first place."

src/Video/video.tsx
samjgorman: Great idea! Love how you reduced this to a much simpler state. I've implemented something quite similar to this from this feedback!

src/Video/video.tsx
samjgorman: With our constants PR from @kanyesthaker , this was an easy fix by adding additional colors to the `Colors` constant!

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+    updateVideo();
+  }, [videoOn]);
+
+  async function handleVideo() {
+    setVideoOn(!videoOn);
+  }
+
+  async function handleAudio() {
+    setAudioOn(!audioOn);
+  }
```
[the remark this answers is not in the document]
samjgorman: Nice catch here, applied

src/Video/video.tsx
samjgorman: I love this! Great insight to simplify this in such an elegant way. I think this is the sort of thing that I'll be able to catch & refactor on a second pass before a PR next round

src/Video/video.tsx
samjgorman: Thanks for the feedback here!  Still new to using `enums` and this is great. 

I'd note that at least in Typescript, `string enums` are supported.  They're useful when debugging, or when you'd like  a meaningful and readable value when your code runs, independent of the name of the enum member itself.
https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

Ex)
```typescript
enum Direction {
  Up = "UP",
  Down = "DOWN",
  Left = "LEFT",
  Right = "RIGHT",
}
```

That said, I think it makes sense to refactor based on the suggestion you gave!  So you should see that reflected in the PR.

src/Video/video.tsx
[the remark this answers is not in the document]
samjgorman: +1!

More reading:

https://reactwithhooks.netlify.app/docs/hooks-state.html

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+      };
+    }
+    return newPermissions;
+  }
+
+  useEffect(() => {
+    async function getMedia() {
+      try {
+        let permissions = configurePermissions();
+        startMedia(permissions);
```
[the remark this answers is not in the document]
samjgorman: Now awaiting `startMedia()`, but do want to highlight one aspect of `useEffect` that requires you to call an `async` function without `awaiting`.  This is not the issue you highlighted, but explains why I do not await the `getMedia` when invoking in `useEffect`

Excerpts from: 
https://devtrium.com/posts/async-functions-useeffect

The issue here is that the first argument of useEffect is supposed to be a function that returns either nothing (undefined) or a function (to clean up side effects). But an async function returns a Promise, which can't be called as a function! It's simply not what the useEffect hook expects for its first argument.

· · ·
// ❌ don't do this
useEffect(async () => {
  const data = await fetchData();
}, [fetchData])
· · ·

Rather, write the `async` function in the `useEffect` or use the `useCallback` hook!  Just be sure to have any logic that depends on the data called in the async function handled within this function.

src/Video/video.tsx
```diff
@@ -0,0 +1,208 @@
...
+  useEffect(() => {
+    async function updateAudio() {
+      if (!audioOn) {
+        endMedia(VideoAudioOptions.audio);
+      } else {
+        let permissions = configurePermissions();
+        endMedia(VideoAudioOptions.all);
+        startMedia(permissions);
+      }
+    }
```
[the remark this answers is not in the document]
samjgorman: This needs to be `async`! See my above comment about awaiting in a `useEffect`

samjgorman:
Okay @shalins @kanyesthaker, ready for rereview!  Pretty comprehensive overhaul for readibility & code quality with a couple project QoL additions I'll highlight below.


* Common error states are caught in `startMedia`, and an appropriate error message is rendered in a new `Dialog` component.
* All or most components in the project use ES6 function syntax.

src/UI/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { DialogContainer, DialogText } from './DialogStyled';
+
+interface DialogProps {
```
kanyesthaker, line 4: Should we have a separate "props" file to define component props in?

src/UI/Dialog/DialogStyled.tsx
```diff
@@ -0,0 +1,18 @@
+import styled from 'styled-components';
```
kanyesthaker, line 1: Do we need a new `xStyled.tsx` file for each component? or should we have a universal style file/style directory? also, why is this a tsx file if we're not rendering any tsx components?

src/UI/Display/Display.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { DisplayConstants } from '../../constants';
+import VideoPlayer from '../Video/VideoPlayer';
+
+const Display = () => {
+  return (
+    <>
```
kanyesthaker, line 7: does this need to be nested inside an empty component? Why can't we just return the `<Videoplayer />` component directly?

src/UI/Video/VideoPlayer.tsx
```diff
@@ -0,0 +1,209 @@
...
+ *  The stateful component uses a ref to the video element to manage toggling video & audio.
+ * @param props VideoProps
+ * @returns <VideoPlayer>
+ */
+const VideoPlayer = (props: VideoProps) => {
+  const [audioEnabled, setAudioEnabled] = useState(true);
+  const [videoEnabled, setVideoEnabled] = useState(true);
+  const [errMessageOnUserMedia, setErrMessageOnUserMedia] = useState('');
+
+  let videoRef = useRef<HTMLVideoElement | null>(null);
```
kanyesthaker, line 63: why the optional null?

src/UI/Video/VideoPlayer.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  /**
+   * endMedia obtains a ref to the current MediaStream if it exists,
+   * then iterates through each track of this stream to toggle off audio, video, or all.
+   * @param selectedTrackKind user selected track kind to toggle
+   */
+  const endMedia = (selectedTrackKind: AVOptions) => {
+    if (videoRef.current?.srcObject) {
```
kanyesthaker, line 71: much cleaner!

src/UI/Video/VideoPlayer.tsx
```diff
@@ -0,0 +1,209 @@
...
+    }
+  };
+
+  /*
+   * startMedia starts the stream with a call to navigator.mediaDevices.getUserMedia.
+   * On error, sets an error state and toggles state to render a Dialog component.
+   */
+  const startMedia = async (permissions: MediaStreamConstraints) => {
+    try {
+      let stream = await navigator.mediaDevices.getUserMedia(permissions);
```
kanyesthaker, line 92: good

src/UI/Video/VideoPlayer.tsx
```diff
@@ -0,0 +1,209 @@
...
+      if (videoRef.current) {
+        videoRef.current.srcObject = stream;
+        videoRef.current.play();
+      }
+    } catch (err: any) {
+      if (err.name in AVErrorMessages) {
+        setErrMessageOnUserMedia(AVErrorMessages[err.name]);
+      } else {
+        setErrMessageOnUserMedia(
+          AVErrorMessages[AVErrors[AVErrors.OtherError]]
```
kanyesthaker, line 102: Can we use `AVErrorMessages[OtherError]`? why do we need the `AVErrors` enum at all?

src/UI/Video/VideoPlayer.tsx
```diff
@@ -0,0 +1,209 @@
...
+        );
+      }
+    }
+  };
+
+  /*
+   * configurePermissions sets video & audio permissions for calls to startMedia.
+   */
+  const configurePermissions = () => {
```
kanyesthaker, line 111: all the below functions that are checking for/waiting for permissions seem to be very similar. Can we modularize?

src/UI/Video/VideoPlayer.tsx
```diff
@@ -0,0 +1,209 @@
...
+            ref={videoRef}
+            height={props.height}
+            width={props.width}
+            autoPlay
+          ></Video>
+
+          <IconContainer>
+            <Icon
+              onClick={toggleAudio}
+              icon={audioEnabled ? faMicrophone : faMicrophoneSlash}
```
kanyesthaker, line 194: good use of ternaries

src/UI/Video/VideoPlayerStyled.ts
```diff
@@ -0,0 +1,28 @@
+import styled from 'styled-components';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+export const VideoContainer = styled.div`
```
kanyesthaker, line 4: should we be using constants throughout this file? i.e. for video default height, background color, etc.

kanyesthaker: also, see my earlier question on styled components

src/World/Player/Player.tsx
```diff
@@ -1,20 +1,20 @@
...
 
-export default function Player(props: SphereProps) {
-  /**
-   * Defines a custom player object (a default Cannon.js sphere and handles player movement)
-   * Movement works by examining keypresses and updating the player's velocity (handled by Cannon.js)
-   * according to keypress logic defined in ./playerMovementControls.ts.
-   */
-  const { forward, backward, left, right, jump } = playerMovementControls();
+/**
+ * Defines a custom player object (a default Cannon.js sphere and handles player movement)
```
kanyesthaker, line 12:
```suggestion
 * Defines a custom player object (a default Cannon.js sphere) and handles player movement
```

src/World/Player/PlayerMovementControls.tsx
```diff
@@ -1,12 +1,13 @@
...
-  /**
-   * Handles keypresses, namely how to emit the appropriate translation
-   * from keypresses into player directions, which are then fed to the main
-   * Player.tsx file to handle the conversio from directions (booleans) to
-   * Cannon.js movement vectors
-   */
+/**
+ * Handles keypresses, namely how to emit the appropriate translation
+ * from keypresses into player directions, which are then fed to the main
+ * Player.tsx file to handle the conversio from directions (booleans) to
```
kanyesthaker, line 7:
```suggestion
 * Player.tsx file to handle the conversion from directions (booleans) to
```

src/UI/Video/VideoPlayer.tsx
kanyesthaker: for example, is there a reason to call the configurePermissions function (which involves creating a new object each call) versus having a high level permissions object, and then simply editing the properties of that object?

src/UI/Video/VideoPlayer.tsx
```diff
@@ -0,0 +1,209 @@
...
+      let permissions = configurePermissions();
+      await startMedia(permissions);
+    };
+    getMedia();
+  }, []);
+
+  /*
+   * On change of audioEnabled state, ends or restarts audio media.
+   */
+  useEffect(() => {
```
kanyesthaker, line 133: i think you can combine both the audio and video perm hooks into a single hook

kanyesthaker (commented):
left some more comments, mostly positive w/ a few questions. Small places to optimize/clean up, but just about ready for merge.

src/UI/Dialog/DialogStyled.tsx
samjgorman: I would prefer to have separate files for each component, and at least in the smaller projects I've worked on, typically see that as convention. Also,   Styled-components eliminates the possibility of a namespace collision for a CSS class, so shouldn't need to worry about accidently naming duplicate classes in this split file approach.

This can be a TS file though! i'll update

src/UI/Display/Display.tsx
samjgorman: We can return directly! But I rendered this within a [Fragment](https://reactjs.org/docs/fragments.html)  because I anticipate rendering additional UI components within the Display component in the near future...

src/UI/Dialog/Dialog.tsx
kanyesthaker: ^ actually, the more i think about it the more i like this idea -- just like we have a constants file, we could keep all component props in their own file -- this might be overengineering though, for how simple our project is. Open to either option

samjgorman: I like this idea! I don't think I'd want to go down this route at this moment though.  I looked over our current props and looked for reuse of prop declarations across files, which would be a great signal to move some of these declarations to its own dedicated props file.  I didn't find any notable examples of this, which informed the current thought to hold off on this.  However, as soon as we start to see some kind of consistent reuse across prop types, I'd love to implement this!

src/UI/Video/VideoPlayer.tsx
samjgorman: Nice catch here, this is a vestige of when I was exploring how to handle `null` in a prop type alias. 
UseRef can either be null, or the element that it is referencing, so it's important to handle this case, but a great way to do this that I converged upon was to declare the ref with a default value of `null`, typed to the type of element it plans to receive (in this case, a `HTMLVideoElement`)

So, the code is updated to: 
`  let videoRef = useRef<HTMLVideoElement>(null);
`

src/UI/Video/VideoPlayer.tsx
samjgorman: Just explored this & found that combining this into a single hook introduced a couple subtle bugs related to toggling video & audio states correctly.  I then explored a more verbose approach maintaning separate `updateVideo` and `updateAudio`  in a single useEffect but found that to be less conventional than what the React docs recommend, which is separating concerns with multiple effects.  So, I think I'm okay making the tradeoff of being a bit more verbose in the code for the sake of clarity. Let me know thoughts! 


See more: 
https://reactjs.org/docs/hooks-effect.html#tip-use-multiple-effects-to-separate-concerns

src/UI/Video/VideoPlayer.tsx
samjgorman: Nice eye to simplify this. I removed `AVErrors` and now use dot notation to access the value of OtherError. 

ex) `AVErrorMessages.OtherError`

src/UI/Video/VideoPlayerStyled.ts
samjgorman: Good idea! Let's start with a couple lightweight applications of constants for colors and sizes, and down the line open to expand this to fixed margin & padding spacing intervals, font sizes, etc.

kanyesthaker (approved):
lgtm!

PR #15, opened 2022-04-20
Code cleanup: create constants.ts and refactor utils

kanyesthaker:
### Description of Changes
- Rename `worldConstants.ts` to `constants.ts` and add constants
- Generalize `getS3Path` to `resolveS3Path` and allow for greater flexibility/add error handling
- Use constants wherever possible throughout `Scene.tsx` and `Skybox.tsx` 

### Link to Notion ticket
https://www.notion.so/irisapp/Sequester-magic-numbers-strings-in-a-constants-ts-file-67360204854643f2b7b24a298759dac6

### Testing Procedure
- [x] Run and built locally
No other tests required

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,11 @@
+{
+  "printWidth":80,
+  "tabWidth":2,
+  "useTabs":false,
+  "semi":true,
+  "singleQuote":true,
+  "trailingComma":"all",
+  "bracketSpacing":true,
+  "jsxBracketSameLine":true,
+  "arrowParens":"always"
+}
--- a/src/Scene/Scene.tsx
+++ b/src/Scene/Scene.tsx
@@ -9,7 +9,14 @@ import Skybox from './Skybox';
 
 import { Box, Cylinder, Sphere, GLTFModel } from '../Shapes';
 
-import { PhysicsConstants, Plants, Objects, Colors, Dims, Mass } from '../constants';
+import {
+  PhysicsConstants,
+  Plants,
+  Objects,
+  Colors,
+  Dims,
+  Mass,
+} from '../constants';
 
 export default function Scene() {
   /**
@@ -22,8 +29,8 @@ export default function Scene() {
         <PlayerCameraControls />
         <Skybox />
         <ambientLight />
-        <pointLight position={PhysicsConstants.pointLightconstant} />
-        <Physics gravity={PhysicsConstants.gravityConstant}>
+        <pointLight position={PhysicsConstants.pointLight} />
+        <Physics gravity={PhysicsConstants.gravity}>
           <Player />
           <Sphere
             dimensions={[10]}
@@ -78,7 +85,7 @@ export default function Scene() {
             type={'Static'}
             modelScaleFactor={1}
             bboxScaleFactor={0.5}
-            modelPath={Objects.smallBluePlantPotPath}
+            modelPath={Plants.smallBluePlantPotPath}
             collision={false}
           />
 
@@ -113,7 +120,9 @@ export default function Scene() {
           />
 
           <Floor />
-          <gridHelper args={[Dims.floorSize, Dims.floorSize, Colors.black, Colors.grey]} />
+          <gridHelper
+            args={[Dims.floorX, Dims.floorZ, Colors.black, Colors.grey]}
+          />
         </Physics>
       </Canvas>
     </div>
--- a/src/Scene/Scene.tsx
+++ b/src/Scene/Scene.tsx
@@ -9,8 +9,7 @@ import Skybox from './Skybox';
 
 import { Box, Cylinder, Sphere, GLTFModel } from '../Shapes';
 
-import { getS3Path } from '../utils';
-import { GRAVITY_CONSTANT, POINT_LIGHT_CONSTANT } from '../worldConstants';
+import { PhysicsConstants, Plants, Objects, Colors, Dims, Mass } from '../constants';
 
 export default function Scene() {
   /**
@@ -23,63 +22,63 @@ export default function Scene() {
         <PlayerCameraControls />
         <Skybox />
         <ambientLight />
-        <pointLight position={POINT_LIGHT_CONSTANT} />
-        <Physics gravity={GRAVITY_CONSTANT}>
+        <pointLight position={PhysicsConstants.pointLightconstant} />
+        <Physics gravity={PhysicsConstants.gravityConstant}>
           <Player />
           <Sphere
             dimensions={[10]}
-            mass={10}
+            mass={Mass.heavyObject}
             position={[5, 1, -15]}
             type={'Static'}
-            color={'gray'}
+            color={Colors.green}
             collision={true}
           />
 
           <Box
             dimensions={[10, 10, 20]}
-            mass={10}
+            mass={Mass.heavyObject}
             position={[25, 1, -15]}
             type={'Static'}
-            color={'grey'}
+            color={Colors.grey}
             collision={true}
           />
 
           <GLTFModel
-            mass={10}
+            mass={Mass.heavyObject}
             position={[-50, 30, -50]}
             type={'Static'}
             modelScaleFactor={0.001}
             bboxScaleFactor={0.5}
-            modelPath={getS3Path('models/gunship/scene.gltf')}
+            modelPath={Objects.gunshipPath}
             collision={true}
           />
 
           <Cylinder
             dimensions={[10, 10, 5]}
-            mass={10}
+            mass={Mass.heavyObject}
             position={[50, 1, -15]}
             type={'Static'}
-            color={'green'}
+            color={Colors.green}
             collision={false}
           />
 
           <GLTFModel
-            mass={10}
+            mass={Mass.heavyObject}
             position={[5, 0, 50]}
             type={'Static'}
             modelScaleFactor={1}
             bboxScaleFactor={1}
-            modelPath={getS3Path('models/smallBluePlant/scene.gltf')}
+            modelPath={Plants.smallBluePlantPath}
             collision={false}
           />
 
           <GLTFModel
-            mass={10}
+            mass={Mass.heavyObject}
             position={[50, 0, 30]}
             type={'Static'}
             modelScaleFactor={1}
             bboxScaleFactor={0.5}
-            modelPath={getS3Path('models/smallBluePlantPot/scene.gltf')}
+            modelPath={Objects.smallBluePlantPotPath}
             collision={false}
           />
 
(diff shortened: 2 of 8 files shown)


src/constants.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  static readonly green: string = '#008000';
+}
+
+export abstract class Dims {
+  static readonly floorSize: number = 1000;
+}
+
+export abstract class Mass {
+  static readonly heavyObject: number = 10;
```
samjgorman, line 34: Let's go ahead and define this schema now, and can always go back to modify as needed.

An approach to consider:  have a range scaled consistently across 7 sizes from XXS (extra, extra small) to XXL (extra, extra, large).  Drawing inspiration from scaling font sizes on web, such as described here.
https://fontawesome.com/docs/web/style/size

We'll want to consider where to begin the scale & how much to scale up and down by a constant.
To this end, here's a SO post that describes a range between 1 to 10 as generally respecting real physics rules, so some range higher than that may be a reasonable starting point so that we configure jumping & collisions to our needs. For context, player mass is currently set to 15. 
https://stackoverflow.com/questions/16424500/what-would-be-realistic-values-for-gravity-mass-and-contact-material-in-canno

An example scale using 1.5 as a co
xxs: 4.4
xs:  6.5
s:   10
m:  15
l:  22.5
xl:   33.75
xxl:  50

src/constants.ts
```diff
@@ -0,0 +1,49 @@
+import { resolveS3Path } from './utils';
```
samjgorman, line 1: Consider refactoring the constants present in `player.tsx` to this file

src/constants.ts
```diff
@@ -0,0 +1,49 @@
...
+export abstract class Plants {
+  static readonly smallBluePlantPath: string = resolveS3Path('smallBluePlant/scene.gltf', 'model');
+
+  static readonly mushroomTreePath: string = resolveS3Path('mushroomTree/scene.gltf', 'model');
+}
+
+export abstract class Objects {
+  static readonly gunshipPath: string = resolveS3Path('gunship/scene.gltf', 'model');
+
+  static readonly smallBluePlantPotPath: string = resolveS3Path('smallBluePlantPot/scene.gltf', 'model');
```
samjgorman, line 18: It's safe to move `smallBluePlantPotPath` to the `Plants` class.

src/constants.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export abstract class Colors {
+  static readonly black: string = '#000000';
+
+  static readonly grey: string = '#808080';
+
+  static readonly green: string = '#008000';
+}
+
+export abstract class Dims {
```
samjgorman, line 29: Can you refactor this to an object with `length` and `width` properties rather than a single `floorsize`? To maximize extensibility, we should be able to specify a world size of arbitrary length and width, such as 20 x 2000.

src/constants.ts
```diff
@@ -0,0 +1,49 @@
...
+  static readonly posz: string = resolveS3Path('posz.jpg', 'texture');
+
+  static readonly posy: string = resolveS3Path('posy.jpg', 'texture');
+
+  static readonly negy: string = resolveS3Path('negy.jpg', 'texture');
+
+  static readonly posx: string = resolveS3Path('posx.jpg', 'texture');
+
+  static readonly negx: string = resolveS3Path('negx.jpg', 'texture');
+}
```
samjgorman, line 49: The naming of textures in `SkyboxImages` is confusing to me. Can you explain this more or make it more descriptive?

src/Scene/Skybox.tsx
```diff
@@ -13,17 +13,14 @@ export default function Skybox() {
...
+    SkyboxImages.negx,
   ];
-  const pathStrings = sides.map((side) => {
-    return getS3Path(side);
-  });
-  const texture = loader.load(pathStrings);
+  const texture = loader.load(sides);
   scene.background = texture;
 
   return null;
```
samjgorman, line 26: Why does this function return `null`?

src/constants.ts
```diff
@@ -0,0 +1,49 @@
+import { resolveS3Path } from './utils';
+
+export abstract class PhysicsConstants {
```
samjgorman, line 3: Now that vars are wrapped in the `PhysicsConstants` class, we can drop the constants to their name to increase usability.

ex) 
`PhysicsConstants.gravity` rather than` PhysicsConstants.gravityConstant`

src/constants.ts
```diff
@@ -0,0 +1,49 @@
...
+  static readonly mushroomTreePath: string = resolveS3Path('mushroomTree/scene.gltf', 'model');
+}
+
+export abstract class Objects {
+  static readonly gunshipPath: string = resolveS3Path('gunship/scene.gltf', 'model');
+
+  static readonly smallBluePlantPotPath: string = resolveS3Path('smallBluePlantPot/scene.gltf', 'model');
+}
+
+export abstract class Colors {
```
samjgorman, line 21: Nice work here! I like this, and would like to advocate to use `Colors` as somewhat of a style enforcer for colors that are consistently used across the application.  So for example, I'd like to use semantic naming such as `SpaceBlue`to describe a particular blue used in the application.  So, this usage as an enforcer for a style guide would differ from some large set of colors that could possibly be used in the application but not currently in use. yet.

samjgorman:
@kanyesthaker left a first review! Feel free to take a look.

src/constants.ts
kanyesthaker: I'd prefer adding constants ad hoc vs. pre-empting them, it keeps files brief, but we should save this guide for when it does become relevant

src/constants.ts
kanyesthaker: good idea, forgot about this file

src/constants.ts
kanyesthaker: sure, i'll split this into independent `floorLength` and `floorWidth` constants

src/constants.ts
kanyesthaker: these refer to where these textures are mapped onto the skybox (they correspond to 3d coordinate axes). Can add a comment to `Skybox.tsx` that explains

src/Scene/Skybox.tsx
kanyesthaker: it's a JSX component and cannot have a `void` return value; it must be a JSX component since it needs to be rendered as a component inside of `<Canvas/>` (this is a fiber convention). There's nothing else to return so it must return null (i'm sure theres an alternative, like putting all of this inside `Scene.tsx`, but I like this reusability)

src/constants.ts
kanyesthaker: yep i agree, we can do this as-needed once we're in the process of finalizing a color palette

src/constants.ts
```diff
@@ -0,0 +1,49 @@
+import { resolveS3Path } from './utils';
+
+export abstract class PhysicsConstants { 
+  static readonly gravityConstant: [number, number, number] = [0, -30, 0];
+
+  static readonly pointLightconstant: [number, number, number] = [10, 10, 10]; 
+}
+
+export abstract class Plants {
+  static readonly smallBluePlantPath: string = resolveS3Path('smallBluePlant/scene.gltf', 'model');
```
samjgorman, line 10: Are you using `Prettier` for code formatting? When running locally, this code block is differentially formatted, although that may be on my end and continues to build a case for a precommit linting hook

kanyesthaker: no, i don't have prettier installed -- could you add your prettier configuration steps to the README so when I set up the linter hooks I know what to be looking at?

samjgorman: Yes, will add!

src/Scene/Scene.tsx
```diff
@@ -9,8 +9,7 @@ import Skybox from './Skybox';
 
 import { Box, Cylinder, Sphere, GLTFModel } from '../Shapes';
 
-import { getS3Path } from '../utils';
-import { GRAVITY_CONSTANT, POINT_LIGHT_CONSTANT } from '../worldConstants';
+import { PhysicsConstants, Plants, Objects, Colors, Dims, Mass } from '../constants';
```
samjgorman, line 12: I don't have a strong preference now, but here's an alternative approach to doing this. This may be even more relevant as the list of classes in `constants.ts `grows

`import * as constants from '../constants';`

Use: 
`constants.Mass.gravity`, etc

src/constants.ts
samjgorman: A comment would be great!

src/Scene/Skybox.tsx
samjgorman: Okay, let's keep this. I looked into whether it made more sense to return a `React.Fragment` rather than `null`, but I think null is the way to go.
https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-useless-fragment.md

src/constants.ts
samjgorman: Added!

src/Scene/Scene.tsx
kanyesthaker: thanks for the tip! I discovered that the AirBnB style guide strongly recommends against using wildcard imports: https://airbnb.io/javascript/#modules--no-wildcard

kanyesthaker:
@shalins @samjgorman would love another rev!

src/Scene/Scene.tsx
samjgorman: Thanks for digging into this! Sounds good to me.

.prettierrc
```diff
@@ -0,0 +1,11 @@
+{
```
samjgorman, line 1: Nice addition here to help standardize our prettier config!

src/Scene/Skybox.tsx
```diff
@@ -1,29 +1,28 @@
 import { useThree } from '@react-three/fiber';
 import { CubeTextureLoader } from 'three';
-import { getS3Path } from '../utils';
+import { SkyboxImages } from '../constants';
 
+/**
```
samjgorman, line 5: Thanks for moving the placement of this function comment!

src/constants.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  static readonly negy: string = resolveS3Path('negy.jpg', 'texture');
+
+  static readonly posx: string = resolveS3Path('posx.jpg', 'texture');
+
+  static readonly negx: string = resolveS3Path('negx.jpg', 'texture');
+}
+
+export abstract class PlayerConstants {
+  static readonly speed: number = 10;
```
samjgorman, line 70: Nit:  Typescript can infer that this is a number, so it isn't necessary to add a type annotation in simple cases like this. 

Read more: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-annotations-on-variables

samjgorman:
> @shalins @samjgorman would love another rev!

Reviewed @kanyesthaker ! I left a minor comment or two related to unneeded type annotations.  After you address these, you're good to merge!

src/constants.ts
kanyesthaker: makes sense -- i was thinking of this more from a developer pov (i.e. have the type listed explicitly to force us to always use that type when we change the constant)

samjgorman: Ok! Decision here is left to you-- just wanted to make sure you were aware of how Typescript recommended handled type annotations in cases like these!

samjgorman (approved):
LGTM!

PR #17, opened 2022-04-27
feat: add linting bash script and pre-commit hook

kanyesthaker:
### Description of Changes
Add pre-commit hook to enforce linting, prettier, and import sorting via Husky

_example of successful linting on commit_
image

_example of failed linting on commit_
image

_example of running hook_
image

### Link to Notion ticket
https://www.notion.so/irisapp/Add-ESLint-Prettier-pre-commit-git-hook-a928823195384e35901ac5456467919c

### Testing Procedure
- [x] Test on files with unsorted imports
- [x] Test on files that fail ESLint and Prettier conventions
- [x] Test --no-verify flag to skip running hook on commit


samjgorman:
It may be helpful to include a note in the PR that in order to use & test Husky, you'll need to install with `npm i` first.

package.json
```diff
@@ -59,10 +63,37 @@
...
+    "jsxBracketSameLine": true,
+    "arrowParens": "always"
+  },
+  "importSort": {
+    ".ts, .tsx": {
+      "parser": "import-sort-parser-typescript",
+      "style": "eslint"
+    }
+  },
+  "lint-staged": {
```
samjgorman, line 91: Nit:  What do you think about including documentation on what each command does?   It's not clear to me what       `pretty-quick --staged` does without more context.

Given that JSON does not support comments, I did some research into when people document code within their` package.json`.  I haven't found a standard solution, but I did find two approaches: inline documentation or creating a separate `package.md`file.

Inline documentation:
https://stackoverflow.com/questions/14221579/how-do-i-add-comments-to-package-json-for-npm-install

Separate file:
https://spin.atomicobject.com/2019/05/20/document-package-json/

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
```
samjgorman, line 2: I'd like to see a functional description of the pre-commit hook, the flags that may be invoked & what they do, such as `no-verify` for ex.  Perhaps can do this as a top-level docstring in this file?

package.json
kanyesthaker: since JSON doesn't support comments, I'm reluctant to hack inline comments into the file -- happy to write a separate markdown file to explain packages

.husky/pre-commit
kanyesthaker: since this is package-level and not something we're implementing ourselves, i think it's best to add this to the README.md but to not include it in the scripts (esp since this is a generated file)

samjgorman:
What do you think about including a pre-commit hook to enforce commit message consistency?  This may help us enforce the `type: brief description` commit convention & make future code reviews easier. 

We can do this relatively simply using [`commitlint`](https://github.com/conventional-changelog/commitlint) with Husky.  Commitment enforces the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) commit style by default, but supports an arbitrarily specified set of rules if we'd like to provide our own. The biggest change between what we're currently doing and conventional commits is that the latter enforces including one of a set of standardized verbs like fix, feat, chore, build, etc, then describing the scope that you are working in. For ex) `feat(shapes lib): Introduce new transparent prop type`. I'm okay moving to this style, but also curious to hear if there's a certain style explicitly enforced at Snorkel or Matician beyond what we're currently doing. @shalins @kanyesthaker 

Here's a tutorial to reference, but the docs are likely the most helpful resource for this.
https://danielcorcoranssql.wordpress.com/2021/06/27/setting-up-commitlint-and-husky-with-a-new-project/

samjgorman:
Okay, I've done a first review @kanyesthaker!  Overall solid work, this review was more lightweight given the smaller nature of the PR, but would be excited to potentially introduce a commit message linting hook as well.

.husky/pre-commit
samjgorman: Sounds reasonable!

package.json
samjgorman: Great! Was leaning this direction as well...

kanyesthaker:
> What do you think about including a pre-commit hook to enforce commit message consistency? This may help us enforce the `type: brief description` commit convention & make future code reviews easier.

don't have a strong opinion, snorkel doesn't do this but i can see how it is helpful. @shalins defer

README.md
```diff
@@ -1,95 +1,17 @@
...
+## Commits
 
-## Learn More
+Commits should be named with topic: description. For example, README: updated description. This makes it easier to view when viewing `git log`s.
 
-You can learn more in the [Create React App documentation](https://facebook.github.io/create-react-app/docs/getting-started).
+## Linting
 
-To learn React, check out the [React documentation](https://reactjs.org/).
+Linting is automatically run whenever you `git commit`. Linting does not fix issues by default. Linting on commit will only run the linters (`eslint` and `prettier`) on files staged for commit. In order to run linting on all files, use the `./spacebar` shell script with parameter `lint`. To run linting on all files and fix errors, run `./spacebar lint --fix`. To ignore linting on a commit, run `git commit` with the `--no-verify` flag. Do not close a pull request until all modified files have been linted and fixed.
```
samjgorman, line 17: Thanks for the description here...

src/World/Player/Player.tsx
```diff
@@ -1,12 +1,13 @@
 import { SphereProps, useSphere } from '@react-three/cannon';
 import { useFrame, useThree } from '@react-three/fiber';
 import React, { useEffect, useRef } from 'react';
+import { Vector3 } from 'three';
+
+import { PlayerPosition, PlayerVelocity } from '../../allTypes';
```
samjgorman, line 6: Love seeing the linter in action!

spacebar
```diff
@@ -0,0 +1,83 @@
...
+SpaceBar Shell Utility
+Usage:
+    ./spacebar lint [--fix|-f]
+    ./spacebar --help | -h
+
+Options:
+    -h --help       Show the help menu.
+    -v --verbose    Print bash trace.
+    lint            Run linters on all files. Run with --fix to fix.
+    deploy          [TODO]
```
samjgorman, line 27: If we're leaving TODOs in the code, can you assign yourself to the TODO with something like TODO(Kanyes)? Think of it as our equivalent of a JIRA ticket to make sure we don't leave TODOs in the code that won't be completed...

spacebar
```diff
@@ -0,0 +1,83 @@
...
+    echo 2>&1 Error: "$*"
+    echo "Use the flag -h/--help for more information."
+    exit 1;
+}
+
+function usage () {
+cat<<"EOF"
+SpaceBar Shell Utility
+Usage:
+    ./spacebar lint [--fix|-f]
```
samjgorman, line 20: Appreciate this convention, and nice work including the fix flag

spacebar
```diff
@@ -0,0 +1,83 @@
...
+    -v --verbose    Print bash trace.
+    lint            Run linters on all files. Run with --fix to fix.
+    deploy          [TODO]
+    build           [TODO]
+EOF
+exit 0
+}
+
+main_param=''
+positionals=()
```
samjgorman, line 34: Can we rename this for clarity?` positional_args `for ex?

spacebar
```diff
@@ -0,0 +1,83 @@
...
+        -f|--fix)
+            fix_flag=1
+            shift
+            ;;
+        -v|--verbose)
+            set -x
+            shift
+            ;;
+        -h|--help)
+            usage
```
samjgorman, line 51: rename to `usage_help` for consistency? Or just `help`?

spacebar
```diff
@@ -0,0 +1,83 @@
...
+            ;;
+    esac
+done
+
+set -- ${positionals[@]}
+
+# Extract main positional arg
+if [[ "$#" -eq 0 ]]; then
+   die "Must supply one of [lint] if help flag not set. " 
+elif [[ "$1" =~ ^(lint)$ ]]; then
```
samjgorman, line 66: Is matching on a regex the best way to do this?

spacebar
```diff
@@ -0,0 +1,83 @@
...
+}
+
+function die () {
+    echo 2>&1 Error: "$*"
+    echo "Use the flag -h/--help for more information."
+    exit 1;
+}
+
+function usage () {
+cat<<"EOF"
```
samjgorman, line 17: Why is `cat`ing EOF necessary?

samjgorman (commented):
Overall, great work introducing a linter to the project-- running a command line `spacebar lint --fix` is intuitive and will add needed structure to our project. I'll defer to @shalins for more granular feedback on how your bash script is written, but I did raise questions for certain choices I'm looking forward to hearing your thoughts on...

spacebar
kanyesthaker: https://en.wikipedia.org/wiki/Here_document#Unix_shells
Tells the shell to capture the entire multi-line string (otherwise would stop at newline)

spacebar
kanyesthaker: allows it to be extensible; in the future we can simply do 
```[[ "$1" =~ ^(lint|build|test)$ ]]; then...```

spacebar
kanyesthaker: curious, why?

kanyesthaker: keeping as-is since `usage()` is the convention https://stackoverflow.com/questions/34567826/what-is-usage-in-shell-scripting

samjgorman: Thanks for verifying convention here-- glad to learn it! :)

samjgorman (approved):
LGTM! Excited for this one.

PR #18, opened 2022-05-12
Types: Fix typing bugs & increase code clarity

samjgorman:
### Description of Changes
Small PR to fix QoL bugs related to types and increase code clarity for code involving types in `Player.tsx` & `PlayerCameraControls.tsx`. Future PR will address bugs related to jumping.

### Detailed description
1. Renamed `alltypes.ts` to `types.ts`
2. Refactored types in `types.ts` and cleaned up `Player.tsx`, and `PlayerCameraControls.tsx` for increased code clarity and to resolve new typing issues.
3. Resolved optional props typescript compiler issue from Drei `PointerLockControls` by removing the numerous additional props & including a   // @ts-expect-error comment with a link to the library's open Github issue. 

### Link to Notion ticket

https://www.notion.so/irisapp/Resolve-Drei-optional-props-typing-issue-by-consolidating-custom-types-in-types-ts-4d6c877e4e59432b8b4ef3c5ef37480a

### Testing Procedure

- [x] Replicated expected behavior in local Chrome browser.


src/types.ts
```diff
@@ -0,0 +1,13 @@
+import { Vector3 } from '@react-three/fiber';
+
+/* Type PlayerPosition represents a player's x,y,z coordinates in 3D space.
+   This is a light abstraction for a Vector3 */
+export type PlayerPosition = Vector3 & {
+  x: number;
+  y: number;
+  z: number;
+};
+/* Type PlayerVelocity represents a player's velocity in the x,y,z directions in 3D space.
```
kanyesthaker, line 10: Duplicate comment?

src/Player/playerMovementUtils.ts
```diff
@@ -23,26 +23,28 @@ export function deserialize(data:PlayerPosition) {
 }
 
 /**
- * round is a simple helper funciton that rounds a float 
+ * round is a simple helper funciton that rounds a float
```
kanyesthaker, line 26: function*

src/Player/playerMovementUtils.ts
```diff
@@ -1,20 +1,20 @@
...
  * Returns undefined if error occured.
- * @param data 
+ * @param data
  * @returns string
  */
-export function deserialize(data:PlayerPosition) {
+export function deserialize(data: PlayerPosition) {
   try {
     return JSON.parse(data.toString());
   } catch (e) {
```
kanyesthaker, line 20: If we're going to try-catch let's throw a meaningful error?

src/Player/playerMovementUtils.ts
```diff
@@ -1,20 +1,20 @@
...
 
 /**
  * deserialize is a helper function that is a light wrapper around parsing a JSON string.
  * Returns undefined if error occured.
- * @param data 
+ * @param data
  * @returns string
  */
-export function deserialize(data:PlayerPosition) {
+export function deserialize(data: PlayerPosition) {
```
kanyesthaker, line 17: If we're going to try-catch here let's report a meaningful error?

src/Player/playerMovementUtils.ts
```diff
@@ -1,20 +1,20 @@
-import { PlayerPosition, PlayerVelocity } from '../allTypes';
+import { PlayerPosition, PlayerVelocity } from '../types';
 /**
  * serialize is a helper function that is a light wrapper around stringifying a JSON object.
- * @param data 
+ * @param data
  * @returns string
  */
-export function serialize(data:PlayerPosition):string {
+export function serialize(data: PlayerPosition): string {
```
kanyesthaker, line 7: `deserialize` has error catching (understandably, since JSON serialization is finicky) -- let's maintain consistency here, and add some naive handling to both functions while we're here

src/Player/playerCameraControls.tsx
```diff
@@ -1,48 +1,23 @@
...
+  let controlsRef = useRef<PointerLockControlsProps>(null);
 
   useEffect(() => {
-    if (controls && controls.current) {
-      document.addEventListener('click', () => {
-        controls?.current?.lock();
-      });
-    }
-  }, [controls]);
+    document.addEventListener('click', () => {
```
kanyesthaker, line 15: great cleanup

src/Player/Player.tsx
```diff
@@ -48,14 +48,12 @@ export default function Player(props: SphereProps) {
 
   useFrame(() => {
     let playerCurrentPosition: PlayerPosition = camera.position;
```
kanyesthaker, line 50: curious, should we move `playerCurrentPosition`'s declaration outside the `useFrame()` hook? looks like we're creating a new `PlayerPosition` object every frame (not really a problem since javascript has garbage collection), but might be a good practice to do something like 
```
let playerCurrentPosition: PlayerPosition;
useFrame(()=>{...})
```

src/Player/Player.tsx
```diff
@@ -25,12 +25,12 @@ export default function Player(props: SphereProps) {
...
-  let pastVelocity = new Vector3();
+  let pastPosition: PlayerPosition = new Vector3();
+  let pastVelocity: PlayerVelocity = new Vector3();
 
-  const currentVelocityVector = useRef<PlayerVelocity>({ x: 0, y: 0, z: 0 });
+  const currentVelocityVector = useRef<PlayerVelocity>(new Vector3());
   useEffect(() => {
-    setPlayerRef.velocity.subscribe(playerVelocity => {
+    setPlayerRef.velocity.subscribe((playerVelocity) => {
       currentVelocityVector.current.x = round(
```
kanyesthaker, line 34: can we use our `round` function here?

src/types.ts
```diff
@@ -0,0 +1,13 @@
+import { Vector3 } from '@react-three/fiber';
+
+/* Type PlayerPosition represents a player's x,y,z coordinates in 3D space.
+   This is a light abstraction for a Vector3 */
+export type PlayerPosition = Vector3 & {
```
kanyesthaker, line 5: Out of curiosity, why do we need these types? They seem to be identical and to capture almost the same information, but they add the added complexity of requiring utilities (like `round()`) to accept Union types (i.e. `PlayerPosition | PlayerVelocity`), which feels like it violates some design principles... is there any reason to not just use the standard Vector3 type? Or to define our own custom vector type?

src/types.ts
```diff
@@ -0,0 +1,13 @@
+import { Vector3 } from '@react-three/fiber';
```
kanyesthaker, line 1: As a general rule of thumb, good software design principles enforce using _as few imports as possible_ in areas of the codebase which are _accessed the most_. In this case, the import 
`import { Vector3 } from '@react-three/fiber';`
loads the entire React Three Fiber library, and only adds the scope of the Vector3 component to the current file. While this is technically OK in this project, since all of Fiber is loaded into memory anyway, this is an anti-pattern; we're importing from `types.ts` across nearly every file in the project, and we can imagine that if this file were shared across multiple projects we would quickly begin to see issues. 

As per my other comments, it seems as though the types built on top of Vector3 are just for naming convenience, so I question whether we need to be doing this; even if we decide we want the abstraction, I feel like it is better practice to define a type as simple as Vector3 from scratch.

kanyesthaker: want to point out that your replacement of `import { useThree } from '@react-three/fiber'` with a standard `useRef` hook is a great example of the right way to use this principle! (though again, in this particular case it is more of a stylistic choice since drei loads fiber anyway).

src/Player/playerMovementUtils.ts
```diff
@@ -23,26 +23,28 @@ export function deserialize(data:PlayerPosition) {
...
- * @param vector 
- * @param numToRound 
+ * @param vector
+ * @param numToRound
  * @returns vector
  */
-export function roundEntriesInVector(vector:PlayerPosition | PlayerVelocity, numToRound:number){
-
+export function roundEntriesInVector(
+  vector: PlayerPosition | PlayerVelocity,
```
kanyesthaker, line 43: see other comments below

src/Player/Player.tsx
```diff
@@ -25,12 +25,12 @@ export default function Player(props: SphereProps) {
   const xVector = new Vector3();
   const zVector = new Vector3();
   const newVelocityVector = new Vector3();
-  let pastPosition = new Vector3();
-  let pastVelocity = new Vector3();
+  let pastPosition: PlayerPosition = new Vector3();
```
kanyesthaker, line 28: See comments below

kanyesthaker (commented):
left some comments -- overall, appreciate the cleanup! love the fact that the controls component is now nice and tidy -- also love the use of native react `useRef` hooks over having to use fiber's `useThree` hooks. See my comments in this review about the use of custom types, imports, and whether or not we should be using custom types wrapping over Vector3 at all.

src/Player/playerCameraControls.tsx
samjgorman: Thank you! Typescript docs came in handy!

src/Player/playerMovementUtils.ts
samjgorman: Great idea! Appreciate the consistency

src/Player/playerMovementUtils.ts
samjgorman: Agree! Returning undefined here is not as descriptive or clear as it could be

samjgorman:
Ready for rereview! @kanyesthaker

src/types.ts
samjgorman: Simplified with a `SimpleVector3` type, which can be used when a Vector3 introduces too much overhead in memory to import or complexity in its type checking.  This is especially helpful in the `roundEntriesInVector3` util, where it becomes much simpler to typeguard the vector when indexing into its values.

src/Player/Player.tsx
samjgorman: You're referring to using `roundEntriesInVector3`, correct?

samjgorman: If so, the reason why I didn't use this is that playerVelocity is a `triplet`, which is not a Vector3 or SimpleVector3 that this function accepts. I'll look more into if I can make this function more extensible for a triplet type so that we can cut down on repetitive code

src/Player/Player.tsx
```diff
@@ -28,9 +27,10 @@ export default function Player(props: SphereProps) {
   let pastPosition = new Vector3();
   let pastVelocity = new Vector3();
 
-  const currentVelocityVector = useRef<PlayerVelocity>({ x: 0, y: 0, z: 0 });
+  const currentVelocityVector = useRef(new Vector3());
   useEffect(() => {
-    setPlayerRef.velocity.subscribe(playerVelocity => {
+    setPlayerRef.velocity.subscribe((playerVelocity) => {
+      // currentVelocityVector.current = roundEntriesInVector3(playerVelocity,MathConstants.roundingPrecision);
```
kanyesthaker, line 33: accidentally committed comment?

kanyesthaker, line 33: accidentally committed comment?

src/Player/playerMovementUtils.ts
```diff
@@ -23,26 +29,28 @@ export function deserialize(data:PlayerPosition) {
...
+  vector: SimpleVector3,
+  numToRound: number
+) {
+  for (const [key, value] of Object.entries(vector)) {
+    vector[key as keyof SimpleVector3] = round(value, numToRound);
   }
-  return vector;
-}
\ No newline at end of file
+  return vector as Vector3;
```
kanyesthaker, line 55: is this returning a Vector3, even though the input is a SimpleVector3?

src/types.ts
```diff
@@ -0,0 +1,7 @@
+/* Type SimpleVector3 is a light abstraction for a Vector3. This provides easier type checking & indexing into
+the object compared to a Vector3. */
+export type SimpleVector3 = {
```
kanyesthaker, line 3: From what I'm understanding, a `SimpleVector3` allows us to pass around triplets without the overhead/syntactic messiness incurred with having to use the threejs `Vector3` class. Is this correct?

kanyesthaker (approved):
Left light nits and replied to some comments. Approving ✅

src/Player/playerMovementUtils.ts
kanyesthaker: I still think the best way to handle this is to either
a) not handle the error, which will throw an exception and make the program crash (not recommended in this case -- soft errors should not cause crashes)
b) properly handle the error, which should do some logic to make sure the program does not crash and return the data to the user.

Optional types should, in general, reveal meaningful information, not cover exceptions. Exceptions are thrown for a reason, and it can be dangerous to let them go undetected and handle them as `null`. I think that if there exception handling to be done, it should be done at this level (i.e. don't make an upstream function handle the `undefined`, maybe return an empty JSON object instead?) The upstream function shouldn't have to deal with something failing further down.

src/Player/playerMovementUtils.ts
samjgorman: Now accepts a vector3 for consistency! Internally casts to a SimpleVector3 then casts back to a Vector3 for consistency

PR #19, opened 2022-06-27
feat: procedural terrain generation

### LINKED ISSUE #39: Support procedural terrain generation

kanyesthaker:
### Description of Changes
Procedural terrain generation algorithm based on the plasma-fractal algorithm https://dl.acm.org/doi/10.1145/358523.358553
Reorganized util files into `common/` folder
added math utils to `math.ts` util folder

### Link to Notion ticket
close #39 

### Testing Procedure
image


src/World/Shapes/Plane.tsx
```diff
@@ -21,13 +25,42 @@ interface CustomPlane extends PlaneProps {
...
+      // Need to flatten from A_{x, y} = z to [x, y, z]
+      const terrainParams = {
+        xDim: Dims.floorX,
+        zDim: Dims.floorZ,
+        noiseAmplitude: TerrainConstants.noiseAmplitude,
+        noiseDecay: TerrainConstants.noiseDecay,
+        smooth: true,
+      };
+      const vertices = plasmaFractal(terrainParams).flat();
+      const verticesArray = new Float32Array(vertices.length * 3);
```
samjgorman, line 41: Why does this need to be a Float32Array rather than a new array?

src/World/Shapes/Plane.tsx
```diff
@@ -21,13 +25,42 @@ interface CustomPlane extends PlaneProps {
...
+        }
+      }
+      // Set geometry vertex positions with buffer values
+      if (geomRef.current) {
+        geomRef.current?.setAttribute(
+          'position',
+          new BufferAttribute(verticesArray, 3),
+        );
+        const { position } = geomRef.current?.attributes;
+        position.needsUpdate = true;
```
samjgorman, line 57: Why does this need to be set to true?

src/World/Shapes/Plane.tsx
```diff
@@ -41,15 +74,25 @@ export default function Plane({ type = 'Static', ...props }: CustomPlane) {
...
+        <planeBufferGeometry
+          args={props.dimensions}
+          ref={props.terrain ? geomRef : null}
+        />
+        <meshStandardMaterial
           color={props.color}
           transparent={props.transparent}
           opacity={props.transparent ? 0.0 : 1.0}
           side={DoubleSide}
+          roughness={1.0}
```
samjgorman, line 95: Nice addition!

src/common/constants.ts
```diff
@@ -10,25 +10,33 @@ export abstract class MathConstants {
   static readonly roundingPrecision = 3;
 }
 
+export abstract class TerrainConstants {
+  static readonly noiseAmplitude = 50;
+
+  // Value selected subjectively
+  // decay rate for terrain noise param
+  static readonly noiseDecay = Math.pow(4, -0.6);
```
samjgorman, line 18: 🤔

src/common/terrain.ts
```diff
@@ -0,0 +1,94 @@
...
+  let noiseAmplitude = params.noiseAmplitude;
+
+  const calcDiamondPoints = (): void => {
+    for (let x = 0; x < maxDim; x += stepSize) {
+      for (let y = 0; y < maxDim; y += stepSize) {
+        const topLeft = arr[x][y];
+        const topRight = arr[x + stepSize][y];
+        const botLeft = arr[x][y + stepSize];
+        const botRight = arr[x + stepSize][y + stepSize];
+        const avg = 0.25 * (topLeft + topRight + botLeft + botRight);
```
samjgorman, line 50: Why 0.25 here? Maybe can refactor to constant

src/common/utils/math.ts
```diff
@@ -0,0 +1,96 @@
...
+};
+
+/**
+ * Initializes an empty 2-D array
+ * @param size
+ * @returns
+ */
+const initArray = (size: number) => {
+  var array: Array<Array<number>> = new Array(size);
+  for (var i = 0; i < size; i++) {
```
samjgorman, line 42: Why var? 

And I question why this needs to be a separate function. It's quite short and appears to not be used much across the application. WDYT?

samjgorman (approved):
Nice work @kanyesthaker! Left a few smaller questions, but approving to unblock 🚀

src/World/Shapes/Plane.tsx
kanyesthaker: see https://threejs.org/docs/#api/en/core/BufferAttribute

src/World/Shapes/Plane.tsx
kanyesthaker: "Flag to indicate that this attribute has changed and should be re-sent to the GPU. Set this to true when you modify the value of the array." https://threejs.org/docs/#api/en/core/BufferAttribute

src/common/terrain.ts
kanyesthaker: taking the average of 4 values! can write a docstring for each subfunction

src/common/utils/math.ts
kanyesthaker: good point. doesn't need to really be its own separate function, i'll move it into the main one. also doesn't need to be a `var`

src/World/Shapes/Plane.tsx
kanyesthaker: will add comment

src/World/Shapes/Plane.tsx
kanyesthaker: will add comment

src/common/utils/math.ts
```diff
@@ -0,0 +1,96 @@
...
+    mid = array2D[wrapArray(-1)][j] / 3;
+    bot = array2D[0][j] / 3;
+    for (i = 0; i < array2D.length; i++) {
+      top = mid;
+      mid = bot;
+      bot = array2D[wrapArray(i + 1)][j] / 3;
+      array2D[i][j] = top + mid + bot;
+    }
+  }
+};
```
shalins, line 88: You can just combine this into one double for loop.

```suggestion
  for (var i = 0; i < array2D.length; i++) {
	// For rows
    left = 0;
    right = array2D[i][0] / 3;
    mid_row = array2D[i][wrapArray(-1)] / 3;

	// For cols
    top = 0;
    bot = array2D[0][j] / 3;
    mid_col = array2D[wrapArray(-1)][j] / 3;

    for (var j = 0; j < array2D.length; j++) {
  	  // Apply 1-D moving average to each row
      left = mid;
      mid_row = right;
      right = array2D[i][wrapArray(j + 1)] / 3;
      array2D[i][j] = left + mid + right;

	  // Apply 1-D moving average to each column
	  top = mid;
      mid_col = bot;
      bot = array2D[wrapArray(j + 1)][i] / 3;
      array2D[j][i] = top + mid + bot;
    }
  }
```

src/common/terrain.ts
```diff
@@ -0,0 +1,101 @@
...
+   * 3 - - - 4
+   */
+  const calcDiamondPoints = (): void => {
+    for (let x = 0; x < maxDim; x += stepSize) {
+      for (let y = 0; y < maxDim; y += stepSize) {
+        const topLeft = arr[x][y];
+        const topRight = arr[x + stepSize][y];
+        const botLeft = arr[x][y + stepSize];
+        const botRight = arr[x + stepSize][y + stepSize];
+        const avg = 0.25 * (topLeft + topRight + botLeft + botRight);
```
shalins, line 47: If you're gonna use `* .25` instead of `/ 4`, you could just use a bit shift here for max optimization. If you're going for clarity, then `/ 4` is clearest, plus it would be consistent with your doc comment where you wrote `(1+2+3+4)/4` :).

src/common/terrain.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * Square step; sets middle point to the
+   * average of 4 side points plus noise
+   * - - 1 - -
+   * - - ↓ - -
+   * 2 → M ← 3 ==> M = (1+2+3+4)/4 + noise
+   * - - ↑ - -
+   * - - 4 - -
+   */
```
shalins, line 62: Big fan of the comments btw! Nice work in clearly explaining diamond-square in a visual way.

src/common/terrain.ts
```diff
@@ -0,0 +1,101 @@
...
+  const calcSquarePoints = (): void => {
+    for (let x = 0; x < maxDim + 1; x += halfStepSize) {
```
shalins, line 64:
```suggestion
    for (let x = 0; x <= maxDim; x += halfStepSize) {
```

src/common/terrain.ts
```diff
@@ -0,0 +1,101 @@
...
+      for (
+        let y = (x + halfStepSize) % stepSize;
+        y < maxDim + 1;
```
shalins, line 67:
```suggestion
        y <= maxDim;
```

src/common/terrain.ts
```diff
@@ -0,0 +1,101 @@
...
+    halfStepSize = stepSize / 2;
+  }
+
+  // Center of the world is at height 0
+  const midpoint = arr[maxDim / 2][maxDim / 2];
+  for (let i = 0; i < arr.length; i++) {
+    for (let j = 0; j < arr.length; j++) {
+      arr[i][j] -= midpoint;
+    }
+  }
```
shalins, line 95: Is there a way to do this in-place in the already multiple passes you do throughout this algorithm? This just feels like an entirely unnecessary pass through all million entries again.

src/common/utils/math.ts
```diff
@@ -0,0 +1,96 @@
...
+  // Apply 1-D moving average to each row
+  for (var i = 0; i < array2D.length; i++) {
+    left = 0;
+    mid = array2D[i][wrapArray(-1)] / 3;
+    right = array2D[i][0] / 3;
+    for (var j = 0; j < array2D.length; j++) {
+      left = mid;
+      mid = right;
+      right = array2D[i][wrapArray(j + 1)] / 3;
+      array2D[i][j] = left + mid + right;
```
shalins, line 72: Is there a reason you're doing the `/ 3` in each of the left, mid, right assignments? I.e. instead of doing `right = array2D[i][wrapArray(j + 1)] / 3;`, can you just do only a lookup `right = array2D[i][wrapArray(j + 1)];` and calculate the average in this step: `array2D[i][j] = (left + mid + right) / 3;`

src/World/Shapes/Plane.tsx
```diff
@@ -1,18 +1,22 @@
...
+import { Dims, TerrainConstants } from '../../common/constants';
+import { plasmaFractal } from '../../common/terrain';
+
 /**
  * CustomPlane extends props
  * for usePlane hook used to render a plane in Cannon.JS's physics engine
  */
 interface CustomPlane extends PlaneProps {
-  dimensions: [number, number];
+  dimensions: [number, number, number, number];
```
shalins, line 14: Why did we do this btw? Can we at least write a comment explaining? It's not immediately obvious to me, especially since in `Floor.tsx` you still just populate the extra dimensions with `floorX` and `floorZ`.

src/World/Shapes/Plane.tsx
```diff
@@ -21,13 +25,44 @@ interface CustomPlane extends PlaneProps {
...
+  const geomRef = useRef<PlaneBufferGeometry>();
 
-  let meshProps = {
-    ref: props.collision ? collisionRef : undefined,
-    rotation: props.rotation,
-    position: props.position,
-  };
+  useEffect(() => {
+    if (props.terrain) {
+      // Need to flatten from A_{x, y} = z to [x, y, z]
```
shalins, line 32: I think you meant for this comment to refer to this line: `const vertices = plasmaFractal(terrainParams).flat();`?

src/common/terrain.ts
```diff
@@ -0,0 +1,101 @@
...
+ * algorithm (Fournier et al 1982) https://dl.acm.org/doi/10.1145/358523.358553
+ * @param params a TerrainParams object
+ * @returns a 2-D array where arr[i][j] is the displacement from zero
+ * to be applied to point (i, j)
+ */
+const plasmaFractal = (params: TerrainParams): Array<Array<number>> => {
+  // Get the max dimension, rounded up to the nearest power of 2
+  const maxDim = math.powRoundUp(2, Math.max(params.xDim, params.zDim));
+  // Instantiate terrain array
+  let arr = math.initArray(maxDim + 1);
```
shalins, line 23: Why + 1 here?

src/common/terrain.ts
```diff
@@ -0,0 +1,101 @@
...
+  arr[0][0] = math.gaussian(0, params.noiseAmplitude);
+  arr[0][maxDim] = math.gaussian(0, params.noiseAmplitude);
+  arr[maxDim][maxDim] = math.gaussian(0, params.noiseAmplitude);
+  arr[maxDim][0] = math.gaussian(0, params.noiseAmplitude);
+  let stepSize = maxDim;
+  let halfStepSize = stepSize / 2;
+  let noiseAmplitude = params.noiseAmplitude;
```
shalins, line 30: Write a comment here explaining that these variables are being declared since they will be modified during the algorithm. Otherwise it's not immediately clear why these are being separately assigned.

shalins (approved):
Really nice work! The way you implemented + commented your code made it very easy to follow what was going on, and I had a ton of fun reading the paper! I'm approving since most of my comments are just nits + minor performance optimizations from an implementation POV. In terms of the actual algorithm, I couldn't find any obvious issues, both from doing multiple passes on the code itself and from testing visually by building the world.

src/common/utils/math.ts
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Uses the Box-Muller transform to generate Gaussian samples
+ * @param mean
+ * @param variance
+ * @returns number
+ */
+const gaussian = (mean: number = 0, variance: number = 1) => {
+  var u1 = 0,
+    u2 = 0;
```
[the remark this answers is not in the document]
kanyesthaker: good eye

src/common/utils/math.ts
```diff
@@ -0,0 +1,96 @@
...
+ * then vertically. This construction has O(1) memory and
+ * O(N) time where N is the number of entries in the array
+ * @param array2D
+ * @returns Array<Array<number>>
+ */
+const boxBlur = (array2D: Array<Array<number>>) => {
+  const wrapArray = (idx: number) => {
+    return mod(idx, array2D.length);
+  };
+  var left, mid, right, top, bot;
```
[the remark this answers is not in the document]
kanyesthaker: huh i thought i changed all vars to lets -- you're right, will address

src/common/utils/math.ts
kanyesthaker: will look into this later, but are you sure this is correct? I'm not sure that this actually replicates behavior; for example, if we have a 3x3:
```
1 2 3
4 5 6
7 8 9
```
iterating through the for-loop once would yield:
```
(1+2+4) 2 3
4       5 6
7       8 9
```
when in reality we want the first entry to be `(1 + 2 + 4 + 5)`?

src/common/terrain.ts
kanyesthaker: yes, you're right, we know `arr[maxDim / 2][maxDim / 2]` from the second we instantiate the array. This code is an artifact of when i was originally setting the mean-height to 0 (instead of making the midpoint 0).

src/common/utils/math.ts
kanyesthaker: i suppose i could, though it's the same number of total divisions

src/World/Shapes/Plane.tsx
kanyesthaker: yes -- at a high level, this is done because we need to set how many segments the floor grid is divided into

src/common/utils/math.ts
shalins: No, I think you're right --  they're not the same. I think the intended behavior is for the rows to be computed first, then you perform the column-wise computation on the result. Whereas when you combine it into one double for loop, you're doing the col-wise computation before the rows have been properly computed. I'm sure you can still put this into one double for loop, but it may take much more time, and not worth the effort.

For context, these are the results with my proposed changes: https://jsfiddle.net/nmdtqe0k/2/, and these are the results from what you originally had: https://jsfiddle.net/wepfdx2g/4/

src/common/terrain.ts
kanyesthaker: the terrain gen array needs to be 1 greater than a power of two for the recursive subdivision to terminate

src/common/utils/math.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+};
+
+export default {
+  boxBlur,
+  gaussian,
+  FlatArray2D,
+  mod,
+  powRoundUp,
+};
```
shalins, line 110: I didn't catch this before but I think this is an anti-pattern here. You're doing a "default" export when you mean to do "named" export. Export default as I understand is pretty much used when you actually have a single thing in a file that you want to have exported by default. For example I'd use `export default useCommunicator;` since that's the "default" thing the user will import from that file. However in a Utils file, I'd use "named" exports since the user would want to import things based on the name. So instead of doing one big `export default ...` at the end, they'd declare each thing exportable. For example: `export const boxBlur = ...`.

src/common/utils/math.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Passes a radius 1 box blur (3x3 kernel) over an array
+ * in-place using the fact that the box blur is a separable
+ * filter by passing a moving average filter horizontally
+ * then vertically. This construction has O(1) memory and
+ * O(N) time where N is the number of entries in the array
+ * @param array2D
+ * @returns Array<Array<number>>
+ */
+const boxBlur = (array2D: FlatArray2D) => {
```
shalins, line 71:
```suggestion
export const boxBlur = (array2D: FlatArray2D) => {
```

src/common/utils/io.ts
```diff
@@ -4,7 +4,7 @@
  * @param fileclass
  * @returns string
  */
-export function resolveS3Path(path: string, fileclass: string) {
+const resolveS3Path = (path: string, fileclass: string) => {
```
shalins, line 7:
```suggestion
export const resolveS3Path = (path: string, fileclass: string) => {
```

src/common/utils/io.ts
```diff
@@ -14,4 +14,8 @@ export function resolveS3Path(path: string, fileclass: string) {
     default:
       throw new Error('fileclass must be one of "texture", "model".');
   }
-}
+};
+
+export default {
+  resolveS3Path,
+};
```
shalins, line 21:
```suggestion
```

src/common/utils/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as io } from './io';
+export { default as math } from './math';
```
shalins, line 2: This whole thing feels incredibly hard to maintain, and a lot more confusing than it needs to be -- I'd highly recommend just switching to named exports. That's what I did for all the networking code and it seemed to simplify things.

src/common/terrain.ts
```diff
@@ -0,0 +1,99 @@
...
+ * to be applied to point (i, j)
+ */
+const plasmaFractal = (
+  params: TerrainParams,
+): InstanceType<typeof math.FlatArray2D> => {
+  // Get the max dimension, rounded up to the nearest power of 2
+  const maxDim = math.powRoundUp(2, Math.max(params.xDim, params.zDim));
+  // Instantiate terrain array (must be 1 greater than a power of 2)
+  let arr = new math.FlatArray2D(maxDim + 1);
+  // let arr = math.initArray(maxDim + 1);
```
shalins, line 26:
```suggestion
```

src/common/terrain.ts
```diff
@@ -0,0 +1,99 @@
...
+ * @param params a TerrainParams object
+ * @returns a 2-D array where arr[i][j] is the displacement from zero
+ * to be applied to point (i, j)
+ */
+const plasmaFractal = (
+  params: TerrainParams,
+): InstanceType<typeof math.FlatArray2D> => {
+  // Get the max dimension, rounded up to the nearest power of 2
+  const maxDim = math.powRoundUp(2, Math.max(params.xDim, params.zDim));
+  // Instantiate terrain array (must be 1 greater than a power of 2)
```
shalins, line 24: I think what may be more useful here is a few-word description of _why_ that's needed for the function to terminate.

src/common/constants.ts
```diff
@@ -56,9 +64,10 @@ export abstract class Colors {
 }
 
 export abstract class Dims {
-  static readonly floorX = 100;
+  // Floor dimensions must be a power of 2 for terrain generation
```
shalins, line 67: Give context?
```suggestion
  // Terrain generation requires floor dims to be pow of 2 so that ...
```

shalins (approved):
Nice work! Left a few more comments, feel free to merge once done :)

src/common/utils/math.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  get(i: number, j: number): number {
+    return this.arr[this.sideLen * i + j];
+  }
+
+  put(i: number, j: number, obj: number): void {
+    this.arr[this.sideLen * i + j] = obj;
+  }
+}
```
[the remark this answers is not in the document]
kanyesthaker: i don't think this qualifies as a `type` though, but yeah this isn't really a "mathematical" class -- i'll think about this

src/common/utils/math.ts
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Uses the Box-Muller transform to generate Gaussian samples
+ * @param mean
+ * @param variance
+ * @returns number
+ */
+const gaussian = (mean: number = 0, variance: number = 1) => {
+  let u1 = 0,
+    u2 = 0;
```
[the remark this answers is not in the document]
kanyesthaker: lol yeah it's the linter

src/common/utils/math.ts
```diff
@@ -0,0 +1,110 @@
...
+ * filter by passing a moving average filter horizontally
+ * then vertically. This construction has O(1) memory and
+ * O(N) time where N is the number of entries in the array
+ * @param array2D
+ * @returns Array<Array<number>>
+ */
+const boxBlur = (array2D: FlatArray2D) => {
+  const wrapArray = (idx: number) => {
+    return mod(idx, array2D.sideLen);
+  };
```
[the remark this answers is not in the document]
kanyesthaker: no, this is a convenience function just to make this function more readable, but i don't think it has use outside of it

src/common/utils/math.ts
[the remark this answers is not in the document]
kanyesthaker: i actually think that this is a class which has been explicitly defined so that math utils work correctly, and as such i don't know if it makes sense to put it in the `types` file which is meant to be accessed for a wider variety of contexts; as such it makes sense to reduce the number of dependencies throughout the process and just keep this here for now. If there's a reason for this to be accessible by non-math functions then we can do another refactor.

PR #23, opened 2022-08-08
linting: ran spacebar linter on the entire codebase

shalins:
### Description of Changes
The current spacebar linting script only runs in the src directory, which made the linter fail on the signaler project, which needs to be a separate node projects. Will merge this after the signaler code gets merged.

### Testing Procedure
Tried running the linter and then commit code from the signaler directory and it worked :)


spacebar
```diff
@@ -72,10 +72,10 @@ case $main_param in
     lint)
         if [[ "$fix_flag" -eq 1 ]]; then
             npx eslint --fix "**/*.{ts,tsx}" --debug
-            npx prettier "src/**/*.{ts,tsx}" --write
+            npx pretty-quick "**/*.{ts,tsx}" --write
```
samjgorman, line 75: Why did we need to do this?

samjgorman (approved):
Looks good to me!  Just left a question

spacebar
shalins: So removing `src` was needed to we can run the linter in the `signaling` directory. Prettier was changed to pretty-quick. IIRC we want to use pretty-quick since that's what we use in our `package.json` as part of the commit hook. When I talked to @kanyesthaker last time, I think he didn't realize pretty-quick had a `--write` flag and so he used prettier instead.

PR #34, opened 2022-08-10
Dedupe files in src to world dir

### LINKED ISSUE #37: Remove duped files

samjgorman:
### Description of Changes

- Dedupe files present in the src directory to live in the `World` dir

### Link to Notion ticket
close #37

### Testing Procedure
Built world locally and verified behavior


shalins (approved):
LGTM! Also out of curiosity, why did we end up deleting from the nested directories and moving them into the outermost `src` directory? Shouldn't it have been the other way around?

samjgorman:
> LGTM! Also out of curiosity, why did we end up deleting from the nested directories and moving them into the outermost `src` directory? Shouldn't it have been the other way around?

Thanks! I don't think we quite did this-  rather I deleted the duped outermost directories within `src` such as `Players` and `Shapes`, as these directories were also present within the `World` directory.

In this iteration of the file structure, we have a UI folder that contains components related to rendering 2D elements, and a World folder that contains components related to rendering the 3D world. So, this PR makes sure that folders such as Player and Shape are within the nested World folder.

shalins:
Ah you're totally right, I didn't realize the kept files were in `World` instead of `src`. Thanks for clarifying. I think you should be good to rebase and merge!

PR #73, opened 2022-08-21
Send coordinate with useCommunicator & render peer coordinate update

### LINKED ISSUE #52: Re-render player objects when coordinates are received from other players

samjgorman:
### Description of Changes
Send coordinates to peer with `useCommunicator` hook 
Rerender a `Peer`component upon receiving updates from the useCommunicator's onMessage callback

### Link to Issue
closes #52 

### Testing Procedure
Manual QA across two instances running on different ports


src/constants.ts
```diff
@@ -8,6 +8,8 @@ export abstract class PhysicsConstants {
 
 export abstract class MathConstants {
   static readonly roundingPrecision = 3;
+
+  static readonly rangeForRandomStartCoordinates = 50;
```
shalins, line 12: Add a comment here?

src/World/Peer/Peer.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+import { Colors, Mass } from '../../constants';
```
shalins, line 3: This is for a later PR but I feel like constants like UI colors and physics concepts like object heaviness should not be in the same const file. Would make more sense if these were split into their relevant files.

src/World/Peer/Peer.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Box } from '../Shapes';
+
+interface PeerProps {
+  peerPosition: string;
+}
+
+/**
+ * Peer renders a Player component upon updates from onMessage
+ */
+const Peer = ({ peerPosition }: PeerProps) => {
```
shalins, line 13:
```suggestion
const Peer = ({ position }: PeerProps) => {
```

src/World/Peer/Peer.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import { Colors, Mass } from '../../constants';
+import { Box } from '../Shapes';
+
+interface PeerProps {
+  peerPosition: string;
+}
+
+/**
+ * Peer renders a Player component upon updates from onMessage
```
shalins, line 11: Maybe you meant to say "Peer renders a Box component" since the code doesn't render a Player? I also don't think it's a good idea to get into the specifics such as naming "onMessage", since if I later changed the callback to "onReceive" in Communicator, now the comment is going to be outdated and incredibly hard to find & fix later.

src/World/Peer/Peer.tsx
```diff
@@ -0,0 +1,32 @@
...
+ */
+const Peer = ({ peerPosition }: PeerProps) => {
+  const data = JSON.parse(peerPosition);
```
shalins, line 14:
```suggestion
  const position = JSON.parse(positionData);
```

src/World/Peer/Peer.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  return (
+    <>
+      {data && (
```
shalins, line 18:
```suggestion
      {position && (
```

src/World/Peer/Peer.tsx
```diff
@@ -0,0 +1,32 @@
...
+        <Box
+          dimensions={[1, 1, 1]}
+          mass={Mass.heavyObject}
+          position={[data.x, data.y, data.z]}
```
shalins, line 22:
```suggestion
          position={[position.x, position.y, position.z]}
```

src/World/Player/Player.tsx
```diff
@@ -1,27 +1,57 @@
...
+import { useSphere } from '@react-three/cannon';
 import { useFrame, useThree } from '@react-three/fiber';
-import React, { useEffect, useRef } from 'react';
+import React from 'react';
+import { useEffect, useRef } from 'react';
 import { Vector3 } from 'three';
 
 import { MathConstants, PlayerConstants } from '../../constants';
+import useCommunicator from '../../network/hooks/useCommunicator';
 import PlayerMovementControls from './PlayerMovementControls';
```
shalins, line 9: Did you run the linter? These don't look ordered.

src/World/Player/Player.tsx
```diff
@@ -1,27 +1,57 @@
...
+
+  const communicator = useCommunicator(
+    roomId,
+    onConnect,
+    onMessage,
+    onDisconnect,
+    onError,
+  );
   const { camera } = useThree();
   const [playerRef, setPlayerRef] = useSphere(() => ({
```
shalins, line 45: Do these really need to be defined inside of the component itself? I feel like you should move these out. For example, no reason to define `generateRandomStartCoordinate` every single time the `Player` component renders. Also I think typescript has a `private` keyword (https://coryrylan.com/blog/private-methods-and-properties-in-typescript-classes), can we impose access control if we move them out?

src/World/Player/Player.tsx
```diff
@@ -1,27 +1,57 @@
...
     mass: PlayerConstants.mass,
-    position: [0, 2, 0],
+    position: [
+      generateRandomStartCoordinate(),
+      2,
```
shalins, line 49: This should be a constant.

src/World/Players/Players.tsx
```diff
@@ -0,0 +1,26 @@
...
+  );
+
+  const onMessage = (message: any) => {
+    setPeerPosition(message);
+  };
+
+  return (
+    <>
+      <Player onMessage={onMessage} />
+      {peerPosition && <Peer peerPosition={peerPosition} />}
```
shalins, line 21: No need to do this check twice. If you're doing this check here, you can remove the `data && <Box ... >` check in the `Peer.tsx` file and just do `return Box<...>`. Since this component tracks state, it makes sense to do the check here.

src/World/Players/Players.tsx
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Players renders a Player and Peer, keeping track of the state for peerPosition
+ */
+const Players = () => {
+  const [peerPosition, setPeerPosition] = useState<string | undefined>(
+    undefined,
+  );
+
+  const onMessage = (message: any) => {
+    setPeerPosition(message);
```
shalins, line 15: Deserialize it here before setting peer position, instead of doing it in the Peer class, that way you go from Network -> Players (where the cleanup happens) -> Peer (which just handles the rendering), instead of splitting the logic arbitrarily across files, which makes it hard for future editors (including yourself) to know if the data has been deserialized or not without having to dig for it.

shalins (dismissed):
Tested it on my local machine, and things seem to be working -- great job! I just left some style and design comments, but for the most part, LGTM!

src/World/Player/Player.tsx
```diff
@@ -1,27 +1,57 @@
...
+  const generateRandomStartCoordinate = () =>
+    Math.floor(Math.random() * MathConstants.rangeForRandomStartCoordinates);
+
+  const communicator = useCommunicator(
+    roomId,
+    onConnect,
+    onMessage,
+    onDisconnect,
+    onError,
+  );
```
shalins, line 43: So I see an issue here, which is that we are creating the `Communicator` class in the `Player` component, which is a child of the `Players` component. However, the way Communicator was designed, we should be declaring it once at the top most component and passing a reference to it around to all the children components. Otherwise, if we move forward with this design, we will have to create a new instance of the `Communicator` class every time we want to call the `.send(...)` function.

shalins:
Ok, so I requested changes because after a second pass, I realized that the Communicator class should be instantiated at the highest level of the component tree. Other than that change, things should be okay!

src/World/Peer/Peer.tsx
samjgorman: Nice catch here! I wouldn't say it would be incredibly hard to find-- you could just do a search for all string matches of onMessage before any change to the naming of the method, which could be good practice when doing any major changes in bigger projects to also update any associated docs that use the older naming.

That said though, definitely agree that it's better at a function-level comment to stay higher level conceptually and I've updated accordingly!

src/World/Peer/Peer.tsx
samjgorman: Thanks for flagging-- perhaps this could get bundled into @kanyesthaker's file restructure PR

src/World/Players/Players.tsx
samjgorman: Great catch here! 🔥  Updated to only conditionally render in `Players.tsx`

src/World/Players/Players.tsx
samjgorman: Good idea here!  Cleaning this up to do this

src/World/Player/Player.tsx
samjgorman: I haven't seen access control as a FE pattern as much, so while it's possible, I think I'll move forward without it for the time being.  _However_, definitely agree that this helper should live in a utils or out of this component, and refactoring!

src/World/Player/Player.tsx
```diff
@@ -1,27 +1,62 @@
...
   const { camera } = useThree();
   const [playerRef, setPlayerRef] = useSphere(() => ({
     mass: PlayerConstants.mass,
-    position: [0, 2, 0],
+    position: [
+      generateRandomStartCoordinate(
```
kanyesthaker, line 49: would make this a helper function for clarity. but optional for now.

kanyesthaker: (i.e. the entire thing, something like `generatePlayerSpawnCoords`

kanyesthaker (approved):
looks good! as discussed, we can defer the reorg (place `useCommunicator` at the top-level) for a future PR

PR #78, opened 2022-09-06
task: restructure project to simplify organization and data flow

### LINKED ISSUE #36: Consistent, clear project structure that respects dataflow patterns

kanyesthaker:
### Description of Changes
Major file restructure (see screenshot). Better reflects dataflow patterns, and more cleanly organizes files. WIP.
image

### Link to Issue
closes #36

### Testing Procedure
N/A


samjgorman (commented):
This is a bit easier reviewed with top-level comments rather than per-line...

Overall nice work!  I don't have any major problems with how you've structured our files.

1. Can you consolidate `allTypes`.ts & `types`into a single types within common?
2. Let's go ahead and prune unused files from our directory. 
      app.test.tsx, server.ts, setupTests.ts are clear candidates.  I've tested that the package.json within src isn't required to build the project successfully-- believe you can remove this as well. 

3. Why did we rename shapes library to models? Wondering if shapes is conceptually clearer
4. When we introduce tests to our project, I'd prefer to move to a format where we have a folder for each component, and then can include tests & styling per component in that folder as well. But for now current format works great!
5. It's a bit odd to me to have this signaler directory at the same level of the src & public directories for the Spacebar client app. What if we refactored to have something like below where all client code is contained in the same directory...

signaler
client
     src
     public
     ....
     
6. Nit: Is it more obvious what the `spacebar` bash script does if it was contained within a `scripts` directory?

samjgorman (approved):
Thanks for addressing changes! Looking 🧨 !

PR #85, opened 2022-09-17
communicator: change hook to not accept callbacks

shalins:
### Description of Changes
Updates the Communicator API to not accept functions in the useCommunicator hook so that it's easier to pass the ref around and listen to event changes in different parts of the code.

### Testing Procedure
Built and tested the world, and it renders players correctly for N=2 peers in a room.


src/World/Player/Player.tsx
```diff
@@ -92,9 +75,7 @@ const Player = ({
 
     if (pastPosition !== playerCurrentPosition) {
       pastPosition = playerCurrentPosition;
-      if (communicator.current) {
-        communicator.current.send(playerCurrentPosition);
-      }
+      communicator.current?.send(playerCurrentPosition);
```
samjgorman, line 78: Nice!

src/network/hooks/useCommunicator.ts
```diff
@@ -1,48 +1,17 @@
...
-  onConnect?: CommunicatorCallback,
-  onMessage?: CommunicatorCallback,
-  onMedia?: CommunicatorCallback,
-  onDisconnect?: CommunicatorCallback,
-  onError?: CommunicatorCallback,
-) => {
+const useCommunicator = (roomID: string) => {
   const communicator = useRef<Communicator>();
 
   useEffect(() => {
```
samjgorman, line 13: What happens when a user inputs an incorrect name for a listener?

samjgorman (approved):
Approving to unblock-- looks good!

PR #93, opened 2022-10-02
signaling: fixed bugs in signaler / communicator

### LINKED ISSUE #92: Fix signaling server crash when multiple WebSocket connections are open for the same peer in the same room.

shalins:
### Description of Changes
As a TLDR, I've fixed the bug in which having a client with multiple tabs open could crash the signaling server. I have also added a hack that temporarily patches the issue where WebSocket connections are dropped when inactive, and require a page refresh from _both_ peers before reconnection is established. This is still not a great solution, however, so it will require a part 2 in which I will do a deeper dive into solving issues that come from multiple browser windows & inactive sessions.

### Link to Issue
Closes #92 

### Testing Procedure
Testing procedure and bug is written about in detail here: https://irisapp.notion.site/Debugging-Signaling-Server-Crash-67acc5f6302f445aa98f53d7662fdffe


src/network/communicator.ts
```diff
@@ -35,11 +35,26 @@ export default class Communicator {
...
     return this.signalingClient.isSocketConnected;
   }
 
+  /** Set whether or not we are connected to the signaling server. */
+  set serverConnected(val: boolean) {
+    this.signalingClient.isSocketConnected = val;
+  }
+
+  /** Reconect to the signaling server. */
+  reConnect() {
```
samjgorman, line 54: Why not `reconnect`?

src/network/communicator.ts
```diff
@@ -35,11 +35,26 @@ export default class Communicator {
     return this.signalingClient.isPeerConnected;
   }
 
+  /** Set whether or not we are connected to the other peer. */
+  set peerConnected(val: boolean) {
```
kanyesthaker, line 39: nit: use more descriptive vars like `isSocketConnected` and `isPeerConnected`?

src/network/signalingClient.ts
```diff
@@ -97,7 +114,7 @@ export default class SignalingClient {
 
   /**
    * Handle when the websocket connection is opened.
-   * @param {any} event The metadata sent with the open event.
+   * @param {Event} event The metadata sent with the open event.
```
samjgorman, line 117: Nice work here!

samjgorman (approved):
Looks solid! 🚢

src/network/signalingClient.ts
```diff
@@ -51,10 +55,23 @@ export default class SignalingClient {
...
+    this._openWebSocketConnection();
+  }
+
+  /** Call when WS connection closes */
+  reConnect() {
+    this._openWebSocketConnection();
+  }
+
+  /** Open a new WebSocket connection and handle events. */
+  _openWebSocketConnection() {
```
kanyesthaker, line 70: nit: is prefixing with an underscore a thing in JS?

kanyesthaker (approved):
some small style nits. otherwise looks good

src/network/signalingClient.ts
shalins: Since JS doesn't have access control, this is mainly for making it obvious this function is private. We use same convention for Python at Matician.

src/network/communicator.ts
shalins: Yeah good point -- this function is removed in next PR so will be addressed.

src/network/communicator.ts
shalins: Good point, will address in next communicator PR.

PR #96, opened 2022-10-31
feat: render video stream onto player avatar (n=1)

### LINKED ISSUE #95: Render AV from communicator onto n=1 other peer avatar

kanyesthaker:
### Description of Changes
Enable peer-to-peer audio and video to render appropriately in the scene. Peer AV now renders directly on the face of the peer player avatar.

TODO (ideally before merge):
- [x] Dynamically grab user permissions (hardcoded as of now)
- [x] Move media functionality into separate utility file
- [x] Handle peer disconnects

### Link to Issue
closes #95

### Testing Procedure
Tested by opening two peers in two separate browser sessions and asserting that both video and audio transmit correctly. 
image

https://user-images.githubusercontent.com/31911175/202144029-891767fe-7093-4047-8c1d-3f445f40f81f.mov


src/scene/players/Player.tsx
```diff
@@ -101,6 +103,27 @@ const Player = ({ communicator }: PlayerProps) => {
...
+    try {
+      // TODO (KT): Configurable permissions
+      let permissions = { audio: true, video: true };
+      let stream = await navigator.mediaDevices.getUserMedia(permissions);
+      if (communicator.current?.peerConnected) {
+        communicator.current?.stream(stream);
+        log.debug(`${Date.now()}: Streaming to other peers`);
+      }
+    } catch (err: any) {
+      console.log(err);
```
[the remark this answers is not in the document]
kanyesthaker: good call

src/scene/players/Player.tsx
```diff
@@ -101,6 +103,27 @@ const Player = ({ communicator }: PlayerProps) => {
...
+    }
+  };
+  useEffect(() => {
+    const getMedia = async () => {
+      await startMedia();
+    };
+    getMedia();
+  }, [communicator.current?.peerConnected]);
```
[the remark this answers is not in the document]
kanyesthaker: oh hmmm i thought i had done this ... i guess i must have git stashed and then not committed -- good catch nonetheless

src/scene/Floor.tsx
```diff
@@ -20,12 +20,12 @@ const Floor = (props: FloorProps) => {
     Dims.floorX,
     Dims.floorZ,
   ];
-  const position: [number, number, number] = process.env.REACT_APP_TERRAIN_FLAG
+  const position: [number, number, number] = process.env.REACT_APP_TERRAIN_FLAG === 'true'
```
samjgorman, line 23: Docstring that explains why this needs to be a string?

src/scene/Floor.tsx
```diff
@@ -20,12 +20,12 @@ const Floor = (props: FloorProps) => {
...
     ? [0, 0, 0]
     : translate([0, 0, 0]);
   const rotation: [number, number, number] = [-Math.PI / 2, 0, 0];
   const floorProps = {
-    color: Colors.grey,
+    color: Colors.white,
```
samjgorman, line 28: Why was this necessary?

src/scene/models/Box.tsx
```diff
@@ -20,7 +21,17 @@ interface CustomBox extends BoxProps {
  * @returns <Box>
  */
 const Box = (props: CustomBox) => {
```
samjgorman, line 23: Destructure props directly rather than declaring props variable at the top level then destructuring?  

Ex)

const Box = ( {mass, dimensions...}: CustomBox)=>{}

src/scene/models/Box.tsx
```diff
@@ -34,20 +45,32 @@ const Box = (props: CustomBox) => {
...
       mass: mass,
       position: position,
+      rotation: rotation,
       type: type,
       onCollide: onCollide,
       ...props,
     }));
   }
+
+  const boxMaterial = texture ? (
```
samjgorman, line 55: Don't know if doing this as a separate variable is any cleaner than just rendering this inline

src/scene/players/Peer.tsx
```diff
@@ -1,26 +1,50 @@
...
+    if (stream) {
+      const videoElement = document.createElement('video') as HTMLVideoElement;
+      videoElement.srcObject = stream;
+      videoElement.play();
+      setVideoTexture(new VideoTexture(videoElement));
+      log.debug(`${Date.now()}: Set video from peer stream`);
+    }
+  }, [stream]);
+
+  const dimensionTriplet: [number, number, number] = [1, 1, 1];
```
samjgorman, line 30: Is this hardcoded because we haven't implemented rotation yet? Could use some help understanding why this is necessary

src/scene/players/Peer.tsx
```diff
@@ -1,26 +1,50 @@
...
+    position.x,
+    position.y,
+    position.z,
+  ];
+  const boxProps = {
+    dimensions: dimensionTriplet,
+    mass: Mass.heavyObject,
+    position: positionTriplet,
+    color: Colors.grey,
+    collision: true,
```
samjgorman, line 41: Given that`collision`is set to `true` here, we'll want to debug why collisions aren't actually being applied to the peer. We'll do that in next PR

src/scene/players/Player.tsx
```diff
@@ -101,6 +103,27 @@ const Player = ({ communicator }: PlayerProps) => {
     pastVelocity.y = currentVelocityVector.current.y;
   });
 
+  // Start media stream if permissions are available
+  const startMedia = async () => {
+    try {
+      // TODO(KT): Configurable permissions
+      let permissions = { audio: true, video: true };
```
samjgorman, line 110: Right now we get user permissions from the 2D video player UI--here's a code pointer.  It seems like a headache to pass state from this videoplayer to the player component here.  Could this be a candidate for a context or Redux style state management mechanism, where the player reads permissions from stored state configured from the 2D video UI?


https://github.com/IrisHub/SpaceBar/pull/14/files#diff-23a944d760dd46506278ca37a8caa536defcbde3225cbc9738017959fcb60ef4R63

src/scene/players/Player.tsx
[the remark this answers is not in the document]
samjgorman: Could be a candidate for an early return within use effect

src/scene/players/Player.tsx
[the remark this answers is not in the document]
samjgorman: Nit: For this message, we're only streaming to one other peer right now. Update the string?

src/scene/players/Peer.tsx
```diff
@@ -1,26 +1,50 @@
...
+  };
+
+  return videoTexture ? (
```
samjgorman, line 44: You can refactor by adding texture to the BoxProps declared above doing something like this! Then, no conditional rendering required

````typescript
const boxProps = {
    dimensions: "hey",
    mass: "hi",
    position: "hello",
    ...(videoTexture ? {} : { texture: "nice" }),
}

````

File: src/scene/players/Player.tsx [hunk 11]
```diff
@@ -101,6 +103,27 @@ const Player = ({ communicator }: PlayerProps) => {
...
+      if (communicator.current?.peerConnected) {
+        communicator.current?.stream(stream);
+        log.debug(`${Date.now()}: Streaming to other peers`);
+      }
+    } catch (err: any) {
+      console.log(err);
+    }
+  };
+  useEffect(() => {
+    const getMedia = async () => {
```
samjgorman, line 121: define this getMedia function outside of the `useEffect`?

src/scene/models/Box.tsx
```diff
@@ -34,20 +45,32 @@ const Box = (props: CustomBox) => {
...
+    <meshBasicMaterial
+      map={texture}
+      transparent={true}
```
samjgorman, line 58: I think we can just write this as `transparent` rather than transparent={true}

src/scene/models/Box.tsx
```diff
@@ -34,20 +45,32 @@ const Box = (props: CustomBox) => {
...
+      opacity={opacity ? opacity : 1.0}
```
samjgorman, line 59: Refactor this to `opacity ?? 1.0`?

samjgorman: And then make the 1.0 a descriptive constant?

samjgorman (commented):
Nice work! Left some comments ⛽

src/scene/players/Player.tsx
kanyesthaker: i believe async useeffects are an antipattern https://devtrium.com/posts/async-functions-useeffect, as are naked async functions? defer

src/scene/models/Box.tsx
kanyesthaker: :o haven't used ?? before -- also i think 0 to 1 is a common range for opacity values; 0 and 1 are also typically exempt from the magic number rule https://www.codereadability.com/magic-numbers/

src/scene/models/Box.tsx
kanyesthaker: yep

src/scene/players/Peer.tsx
kanyesthaker: awesome, love the JS tips, i'll use this

src/scene/players/Player.tsx
kanyesthaker: maybe! I've taken a look at your code pointer. I'd rather not introduce formal state management as much as I can help it, at least not until checkpoint 2. I think I'd just move the files around to make argument passing more straightforward for now

src/scene/players/Peer.tsx
kanyesthaker: yep 🤔 I also just saw this. Ijnvestigating...

src/scene/players/Peer.tsx
kanyesthaker: `dimensionTriplet` is the size of the cube. The reason this is pulled out is because of typing -- if you just pass in `[1, 1, 1]` directly in the props object I'm constructing, typescript casts that as a `[number]` instead of `[number number number]` which fails type validation.

src/scene/models/Box.tsx
kanyesthaker: could be! what do you think?

src/scene/models/Box.tsx
kanyesthaker: good call

src/scene/Floor.tsx
kanyesthaker: I did this so the cubes (which are grey) would stand out from the terrain (which used to be grey). Pretty arbitrary.

src/scene/Floor.tsx
kanyesthaker: I think I'll update the README -- these env vars are always strings. May not make sense to do an inline comment since this is true for all env vars everywhere in the app

src/scene/models/Box.tsx
samjgorman: `?? `is a nullish coalescing operator.  

From MDN docs: 
It returns its right-hand side operand when its left-hand side operand is null and otherwise returns its left-hand side operand.

So in this case, if opacity is null, then we fallback to the default of 1.0

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

samjgorman: Interesting article!  Sounds reasonable to hold off on making this a constant then

src/scene/players/Player.tsx
samjgorman: [You can wrap `getMedia` in a `useCallback` hook and then call that from inside the hook](https://devtrium.com/posts/async-functions-useeffect#what-if-you-need-to-extract-the-function-outside-useeffect).  But given that this is just used once, seems reasonable to leave as is.

src/scene/players/Peer.tsx
samjgorman: Can you make sure this is obvious that `dimensionTriplet` is size of cube?

src/scene/Floor.tsx
samjgorman: Okay! Good call for now then

src/scene/models/Box.tsx
samjgorman: I think this is too shallow to be helpful as a separate variable. Let's put it all back in the return statement?  Abstracting as a separate variable is especially helpful when you have dense or verbose conditional logic that would otherwise obfuscate or distract from the purpose of the component.

README.md
```diff
@@ -19,3 +19,7 @@ Linting is automatically run whenever you `git commit`. Linting does not fix iss
 ## Logging
 
 For your convenience, this library includes a custom environment variable called `REACT_APP_LOG_LEVEL` that lets you view logs at any level. You can just start your node server with `REACT_APP_LOG_LEVEL=xxxxx npm start` where `xxxxx` can be your desired log level. There are 5 levels: Error, Warn, Debug, Info, and Trace, ordered by order of severity. For more information about the framework, you may reference the Javascript [LogLevel framework](https://github.com/pimterry/loglevel).
+
+## Env Flags
+
+Env flags are set by adding the appropriate flag to the root `.env` file, which will toggle behaviors on and off in the application. Feature flags should be prefaced with `REACT_APP`, i.e. `REACT_APP_TERRAIN_FLAG`. The value of a feature flag will always be a string, so truthiness should be checked accordingly. Feature flags will not apply until the server is restarted.
```
samjgorman, line 25: Great!

src/scene/players/Peer.tsx
```diff
@@ -1,27 +1,52 @@
...
+    dimensions: sizeTriplet,
+    mass: Mass.heavyObject,
+    position: positionTriplet,
+    color: Colors.grey,
+    collision: true,
+  };
+
+  const videoBoxProps = {
+    ...boxProps,
+    ...(videoTexture ? { texture: videoTexture } : {}),
```
samjgorman, line 46: Nice!

src/scene/players/Peer.tsx
```diff
@@ -1,27 +1,52 @@
...
+    position.x,
+    position.y,
+    position.z,
+  ];
+  const boxProps = {
+    dimensions: sizeTriplet,
+    mass: Mass.heavyObject,
+    position: positionTriplet,
+    color: Colors.grey,
+    collision: true,
```
samjgorman, line 41: Thought we weren't handling collisions for the peer in this PR?

samjgorman (approved):
Code looks great! Good work with the clean up

src/scene/players/Peer.tsx
kanyesthaker: we aren't, this is bugged. On the roadmap to fix

PR #113, opened 2022-11-19
feat: add spatial audio and remove self-echo

### LINKED ISSUE #54: Create client-side spatial audio functionality

### LINKED ISSUE #97: A player's audio stream should not be sent back to them

### LINKED ISSUE #98: Client-side noise cancelling, echo cancellation

kanyesthaker:
### Description of Changes
- Remove self-audio playback via the UI VideoPlayer
- Add echo cancellation to media stream
- Enable spatial audio
- Rename Peer.tsx and Player.tsx to `PeerPlayer.tsx` and `UserPlayer.tsx`
- Separate `PlayerModel.tsx` component

### Link to Issue
closes #54 
closes #97 
closes #98 

### Testing Procedure
Tested locally -- will test over 1:1 call since it is hard to capture spatial audio functionality via screen recording.


src/common/constants.ts
```diff
@@ -111,3 +113,11 @@ export abstract class DisplayConstants {
 
   static readonly mass = 15;
 }
+
+export abstract class AudioConstants {
+  // Correspond to parameters for the Web Audio Panner Node
+  // https://developer.mozilla.org/en-US/docs/Web/API/PannerNode/PannerNode
```
samjgorman, line 119: Great docstring here

src/scene/models/PlayerModel.tsx
```diff
@@ -0,0 +1,82 @@
...
+ * for useBox hook used to render a box in Cannon.JS's physics engine
+ */
+interface PlayerModelProps extends BoxProps {
+  color?: string;
+  opacity?: number;
+  dimensions: [number, number, number];
+  collision?: boolean;
+  texture?: any;
+  listener?: any;
+  audio?: any;
```
samjgorman, line 16: Nonblocking: For these props typed as `any`, could we consider typing them as `unknown` instead? We introduce more type safety with the unknown type, at the cost of requiring type narrowing before doing most operations.

https://mariusschulz.com/blog/the-unknown-type-in-typescript

samjgorman: In both cases, it may be helpful to have a brief comment justifying your use of permissive typing here

src/scene/players/PeerPlayer.tsx
```diff
@@ -0,0 +1,87 @@
...
+  // If audio strema available, create new positional audio source
+  useEffect(() => {
+    if (audioStream) {
+      const audioElement = document.createElement('audio') as HTMLAudioElement;
+      audioElement.srcObject = audioStream;
+      audioElement.muted = true; // Only want to play from the PositionalAudio source
+      audioRef.current?.setMediaStreamSource(audioElement.srcObject);
+      // https://developer.mozilla.org/en-US/docs/Web/API/PannerNode
+      audioRef.current?.setRefDistance(AudioConstants.AudioRefDistance); // Distance at which audio decay begins
+      audioRef.current?.setDirectionalCone(...AudioConstants.DirectionalCone); // Directional cone outside of which audio is muted
```
samjgorman, line 53: This could benefit from a descriptive helper function name or a comment that takes a step back and describes that this is where we're implementing spatial audio

src/scene/players/PeerPlayer.tsx
```diff
@@ -0,0 +1,87 @@
...
+    color: Colors.grey,
+    collision: true,
+    listener: listener,
+    audio: audioRef,
+  };
+
+  // Set the appropriate props based on what resources are available
+  const videoBoxProps = {
+    ...boxProps,
+    ...(videoTexture ? { texture: videoTexture } : {}),
```
samjgorman, line 75: They grow up so fast 😢

src/scene/players/PeerPlayer.tsx
```diff
@@ -0,0 +1,87 @@
...
+    ...(listener.current ? { listener: listener } : {}),
+    ...(audioRef.current ? { audio: audioRef } : {}),
+  };
+
+  return (
+    <>
```
samjgorman, line 81: Remove this fragment-- it's not necessary. In general, remove any fragments or divs when they aren't required to wrap children

samjgorman (approved):
Left non-blocking comments re: type safety & FE best practice-- otherwise looks awesome! Nice work bundling these changes

src/scene/models/PlayerModel.tsx
kanyesthaker: yeah good point, i was lazy with types here for the first few commits. i'll add in more strict typing before merging

src/scene/players/PeerPlayer.tsx
kanyesthaker: sure, i'll add a comment that mentions that this handles spatial audio

PR #116, opened 2022-11-20
feat: update peer player orientation

### LINKED ISSUE #120: Players should rotate with the mouse

kanyesthaker:
### Description of Changes
Retrieve the quaternion of the player's camera, send it to the peer using communicator, and apply that quaternion to the peer cube object

### Link to Issue
closes #120 

### Testing Procedure
image


src/scene/players/Peer.tsx
```diff
@@ -3,19 +3,19 @@ import React, { useEffect, useState } from 'react';
 import { VideoTexture } from 'three';
 
 import { Colors, Mass } from '../../common/constants';
-import { SimpleVector3 } from '../../common/types';
 import { Box } from '../models';
 
 interface PeerProps {
-  position: SimpleVector3;
+  peerPositionalData: any;
```
samjgorman, line 9: What is the shape of this data? Since it appears to be an object composed of a simple vector peerPosition and array of quaternion values, this seems like it could be somewhat easily typed?

src/scene/players/Players.tsx
```diff
@@ -14,12 +14,12 @@ const Players = () => {
...
 
   /* State variables for peer players */
-  const [peerPosition, setPeerPosition] = useState<SimpleVector3>();
+  const [peerPositionalData, setPeerPositionalData] = useState<SimpleVector3>();
   const [mediaStream, setMediaStream] = useState<MediaStream>();
 
   useEffect(() => {
     communicator.current?.on('message', (data) => {
-      setPeerPosition(JSON.parse(data));
+      setPeerPositionalData(JSON.parse(data));
```
samjgorman, line 22: Because you set quaternion to an array when you encoded `playerPositionalData`, do you need to parse it using .FromArray() when decoding `peerPositionalData`?

And why do you need to convert the quaternion to an array to begin with?

samjgorman (commented):
Looking really good! Left some questions

src/scene/models/Box.tsx
```diff
@@ -24,27 +24,28 @@ const Box = ({
...
+    quaternion: quaternion,
   };
 
   if (collision) {
     [collisionRef] = useBox(() => ({
       args: dimensions,
       mass: mass,
       position: position,
-      rotation: rotation,
+      quaternion: quaternion ?? undefined,
```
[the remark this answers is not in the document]
kanyesthaker: no, you're right we don't need it

src/scene/players/Players.tsx
```diff
@@ -14,12 +14,12 @@ const Players = () => {
   const communicator = useCommunicator(roomId);
 
   /* State variables for peer players */
-  const [peerPosition, setPeerPosition] = useState<SimpleVector3>();
+  const [peerPositionalData, setPeerPositionalData] = useState<SimpleVector3>();
```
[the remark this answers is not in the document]
kanyesthaker: yep totally missed this, will fix

src/scene/players/Peer.tsx
kanyesthaker: yeah we can make this into an actual type, it's quite simple

src/scene/players/Players.tsx
kanyesthaker: good q -- i dont' remember exactly, but i think i was running into issues serializing the quaternion. i'll verify and fix

samjgorman (approved):
Great job here!

PR #117, opened 2022-11-20
feat: single face video

### LINKED ISSUE #105: Peer faces should only be rendered on one side of the cube

kanyesthaker:
### Description of Changes
Makes it so that the videotexture only renders on one face of the cube
This is used by passing an array of materials to the player mesh instead of a single videotexture

### Link to Issue
closes #105 

### Testing Procedure
image


src/scene/models/Box.tsx
```diff
@@ -55,11 +57,18 @@ const Box = ({
       <mesh {...meshProps}>
         <boxBufferGeometry args={dimensions} />
         {texture ? (
-          <meshBasicMaterial
-            map={texture}
-            transparent
-            opacity={opacity ?? 1.0}
```
samjgorman, line 61: Does setting the opacity inhibit the ability to set a video texture on a single face of the box? Curious why this was removed

samjgorman (approved):
LGTM with one question on the opacity prop

src/scene/models/Box.tsx
kanyesthaker: no, there's just no real reason to have an opacity scale applied to a vidoetexture, since it should implicitly be fully opaque. can add it back if we want flexibiliy

Issue #129, opened 2022-12-08
A user must have knowledge of what other peers are in the room, and whether those peers have audio and video enabled


kanyesthaker:

samjgorman:
Let's split these into two tickets, starting with what peers are in the room

kanyesthaker:
Closing this since we have #135

PR #143, opened 2022-12-11
Implement Jotai as state store + indicator component

samjgorman:
### Description of Changes
- Import Jotai as a state store
- Use Jotai to store communicator globally across the app
- Refactor components to read communicator from Jotai 
- Indicator component that reads from jotai state to determine num players in room
- Simplified communicator from a ref of the class to the direct class stored in our state store

In a future PR, will need to properly handle when peers disconnect from the room.  Currently, the indicator does not update after a peer leaves.  I think it makes sense to tackle this along with no longer rendering a peer's cube when they leave the room
https://github.com/IrisHub/SpaceBar/issues/151
https://github.com/IrisHub/SpaceBar/issues/150

### Link to Issue
#129 
### Testing Procedure
Demod to the team + tested across two windows


src/network/communicatorAtomStore.ts
```diff
@@ -0,0 +1,12 @@
...
+import Communicator from './communicator';
+
+/**
+ * Store for globally accessible state of communicator
+ */
+
+export const communicatorAtom = atom<Communicator | null>(null);
+export const isPeerConnected = atom<boolean>(false);
+
+export default null;
```
shalins, line 12:
```suggestion
```
I think we should enforce proper importing of the consts above instead of doing this.

src/network/hooks/useInitializeCommunicator.ts
```diff
@@ -0,0 +1,23 @@
...
+import { useEffect } from 'react';
+
+import Communicator from '../communicator';
+import { communicatorAtom } from '../communicatorAtomStore';
+
+/**
+ * Custom hook to initialize the communictor and attach all
+ * callback functions to the communicator events.
+ * @param {string} roomID The room ID to join.
+ * @returns {CommunicatorRef} A reference to the Communicator.
```
shalins, line 11: I don't think this function returns a `CommunicatorRef`

src/scene/players/Players.tsx
```diff
@@ -23,11 +26,15 @@ const Players = () => {
 
   // Set communicator callbacks
   useEffect(() => {
-    communicator.current?.on('message', (data) => {
+    communicator?.on('message', (data) => {
```
shalins, line 29: Nice side effect of using Jotai!

src/scene/players/Players.tsx
```diff
@@ -36,7 +43,7 @@ const Players = () => {
 
   return (
     <>
-      <UserPlayer communicator={communicator} listener={listenerRef} />
+      <UserPlayer listener={listenerRef} />
```
shalins, line 46: Useful simplification here!

src/App.tsx
```diff
@@ -2,9 +2,13 @@ import React from 'react';
 import { BrowserRouter } from 'react-router-dom';
 
 import Display from './interface/Display';
+import useInitializeCommunicator from './network/hooks/useInitializeCommunicator';
 import Scene from './scene/Scene';
 
 const App = () => {
+  const roomId = 'test_room_id';
+  useInitializeCommunicator(roomId);
```
shalins, line 10: Nice work promoting this to `App`

shalins (changes requested):
Was looking over this PR mostly to understand the changes to the Communicator hook, but had some thoughts along the way so ended up doing a more comprehensive review. Overall, nice work introducing Jotai and moving Communicator to the top level! I'm requesting changes for now because of some of the reasons I've included in the comments below.

src/App.tsx
```diff
@@ -2,9 +2,13 @@ import React from 'react';
 import { BrowserRouter } from 'react-router-dom';
 
 import Display from './interface/Display';
+import useInitializeCommunicator from './network/hooks/useInitializeCommunicator';
```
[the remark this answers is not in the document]
shalins: Update: after reading through the rest of code, I think it makes most sense just to update useCommunicator directly.

src/network/hooks/useInitializeCommunicator.ts
samjgorman: Nice catch!

src/App.tsx
```diff
@@ -2,9 +2,13 @@ import React from 'react';
...
 import Scene from './scene/Scene';
 
 const App = () => {
+  const roomId = 'test_room_id';
```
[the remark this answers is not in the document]
samjgorman: Great idea! Done

src/interface/VideoPlayer.tsx
```diff
@@ -93,6 +93,7 @@ const VideoPlayer = ({ height, width }: VideoProps) => {
       let stream = await navigator.mediaDevices.getUserMedia(permissions);
       if (videoRef.current) {
         videoRef.current.srcObject = new MediaStream(stream.getVideoTracks());
+
```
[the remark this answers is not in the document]
samjgorman: Nice catch-- not needed here!

src/network/hooks/useInitializeCommunicator.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+const useInitializeCommunicator = (roomId?: string) => {
+  const [isCommunicatorInitialized, setIsCommunicatorInitialized] = useAtom(
+    isCommunicatorInitializedAtom,
+  );
+  const setCommunicator = useSetAtom(communicatorAtom);
+
+  if (!roomId) {
+    console.log(
```
shalins, line 24:
```suggestion
    console.warn(
```

src/scene/players/Players.tsx
```diff
@@ -11,32 +10,27 @@ import UserPlayer from './UserPlayer';
...
+    setPeerMovementData(JSON.parse(message));
+  };
 
-    communicator.current?.on('media', (stream) => {
-      log.debug(`${Date.now()}: Peer stream detected`);
-      setVideoStream(new MediaStream(stream.getVideoTracks()));
-      setAudioStream(new MediaStream(stream.getAudioTracks()));
-    });
-  }, []);
+  useCommunicator({ onMedia: onMedia, onMessage: onMessage });
```
shalins, line 29: Cool change!

shalins (approved):

PR #144, opened 2022-12-11
Revert "feat: update peer player orientation"

kanyesthaker:
Reverts IrisHub/SpaceBar#116

Updating the quaternion in this way causes an error in AudioListener where the position vector is set to null on some frames. Needs investigation.


samjgorman (approved):
See you soon!

PR #154, opened 2022-12-23
SFU Server Implementation for Multipeer Communicator

### LINKED ISSUE #152: As discussed, update the Communicator API to better support mulitple peers.
This includes:
* Metadata field with Peer ID's attached to each stream and message.
* Property that tracks the total number of peers in the room.
* Updated variables that keep track of who's connected and who isn't.
* Callback functions for each time a new peer joins or an existing peer leaves.

shalins:
### Description of Changes
This PR enables 3/3 of Communicator!
* Multipeer support for Communicator. For the user, this means that Spacebar will work seamlessly with 3+ peers.
* Converted the Signaling Server into a Signaling + SFU (Selective Forwarding Unit) server, which means that it figures out how to create and maintain both a WebSocket and a P2P connection with each user that uses the website, and properly routes messages between users.
* Lots of new features, including caching messages when the peer has not attached listeners, full backwards compatibility with multitab communication, support for sending messages over both WebSocket and P2P, etc.

The client side implementation is coming soon, which will enable the user to actually see many peers in the world. :)

### Link to Issue
Closes #152 

### Testing Procedure
Tested by trying to use it under various circumstances with various edge cases.


signaler/server.ts
```diff
@@ -16,6 +18,13 @@ export class SignalingServer {
 
   connections: Map<WebSocket, Connection>;
 
+  // Since we're using SimplePeer in a node environment, we need to pass
+  // in the `wrtc` module: https://github.com/feross/simple-peer#in-node
+  // Also, on M1 macs, installing `wrtc` will fail since arm64 binaries are
```
kanyesthaker, line 23: This should only impact local testing, correct? Not when deployed?

signaler/server.ts
```diff
@@ -93,24 +101,32 @@ export class SignalingServer {
...
     // indices, since we could run into races if multiple clients
     // are joining and leaving at the same time.
-    console.log(`Removing peer ${peerID} from room.`);
-    peersInRoom = peersInRoom.filter((peer) => peer.peerID !== peerID);
+    peersInRoom = peersInRoom.filter((peer) => {
+      if (peer.peerID === peerID) {
+        // Remove the peer's stream from all other peers.
+        this._removeStream(roomID, peer.stream);
+
+        // Destroy the simple-peer instance and add a small sleep
```
kanyesthaker, line 109: Where is the "small sleep"?

signaler/server.ts
```diff
@@ -197,71 +201,178 @@ export class SignalingServer {
...
+  }
+
+  /**
+   * Broadcasts a stream to all peers in a room, except the one that
+   * originally sent the stream.
+   * @param {string} roomID ID of the room to broadcast to.
+   * @param {string} peerID ID of the peer that the stream belongs to.
+   * @param {MediaStream} stream Stream to broadcast.
+   */
+  _broadcastStream(roomID: string, peerID: string, stream?: MediaStream) {
```
kanyesthaker, line 312: This is purely out of curiosity -- can these functions (`_broadcastStream` and `_removeStream`) be parallelized? These kinds of fan-out start/stop operations seem like a classic place where you'd want that kind of parallelism. Does that line of thinking make sense (of course, ignoring technical limitations with JS's single-threaded event loop).

signaler/server.ts
```diff
@@ -93,24 +101,32 @@ export class SignalingServer {
     // Delete a peer by `filter`ing instead of using `splice` with
```
kanyesthaker, line 101: Nit: suggest updating this docstring. While it _is_ interesting why we're using filter instead of splice, I think it's more helpful to say something like "Find which peer we're trying to delete by filtering. Then delete the peer's stream from all other peers before destroying the peer itself" (i.e. focus on functionality)

signaler/server.ts
```diff
@@ -143,9 +159,16 @@ export class SignalingServer {
    * @param {WebSocket.RawData} signal Signal message from the peer.
    */
   _handleSignal(roomID: string, peerID: string, signal: WebSocket.RawData) {
-    log.trace('Signal Received: \n', JSON.parse(signal.toString()));
-    const data = createPayload(PayloadType.SIGNAL, signal);
-    this._broadcast(roomID, peerID, data);
+    log.trace(
+      `Signal Received from peer ${peerID}: \n ${JSON.parse(
```
kanyesthaker, line 163: Curious, why do we need to `JSON.parse(signal.toString()` vs. printing the signal directly?

signaler/server.ts
```diff
@@ -197,71 +201,178 @@ export class SignalingServer {
    */
   _createNewPeer(ws: WebSocket, roomID: string, peerID?: string) {
     const peersInRoom = this.rooms.get(roomID).peers;
-    const initiatorsPresent = peersInRoom.find(
-      (peer) => peer.initiator === true,
-    );
     const seenPeer = peersInRoom.find((peer) => peer.peerID === peerID);
 
+    const simplePeer = new SimplePeer({
```
kanyesthaker, line 206: Why don't we need to include ICE candidate config information here (like we currently do in `src/network/peerManager.ts`?

signaler/server.ts
```diff
@@ -197,71 +201,178 @@ export class SignalingServer {
...
+      seenPeer.simplePeer = simplePeer;
+      seenPeer.remoteStreamIds = [];
+      seenPeer.isConnectedP2P = false;
       peer = seenPeer;
-      peer.ws = ws;
     } else {
       peer = {
         roomID: roomID,
         peerID: peerID ? peerID : uuid().toString(),
         // If you're the first peer to join, you're the initiator.
```
kanyesthaker, line 226: Should update this comment if we're changing `initiatior: !initiatorsPresent` --> `initiator: false`

signaler/server.ts
```diff
@@ -197,71 +201,178 @@ export class SignalingServer {
...
+    roomID: string,
+    peerID: string,
+    message: string,
+    p2p: boolean = true,
+  ) {
     this.rooms.get(roomID)?.peers.forEach((peer) => {
-      // Don't send to the client who originally sent the message.
-      if (peer.peerID !== peerID && peer.ws.readyState === WebSocket.OPEN) {
-        peer.ws.send(message);
+      if (peer.peerID !== peerID) {
```
kanyesthaker, line 279: nice!

signaler/server.ts
```diff
@@ -197,71 +201,178 @@ export class SignalingServer {
...
+    }
+  }
+
+  /** Remove a stream from all peers in a room.
+   * @param {string} roomID ID of the room to remove the stream from.
+   * @param {MediaStream} stream Stream to remove.
+   */
+  _removeStream(roomID: string, stream?: MediaStream) {
+    if (stream) {
+      this.rooms.get(roomID)?.peers.forEach((p) => {
```
kanyesthaker, line 329: What happens if 2 peers leave near-simultaneously? Is this loop "threadsafe" (not sure if that's the right word)

signaler/server.ts
```diff
@@ -197,71 +201,178 @@ export class SignalingServer {
...
+   */
+  _subscribeToPeerIDs(roomID: string, currentPeer: Peer) {
+    this.rooms.get(roomID)?.peers.forEach((p) => {
+      if (p.peerID !== currentPeer.peerID) {
+        log.info(
+          `Peer ${currentPeer.peerID} subscribing to peerID: ${p.peerID}`,
+        );
+        const message = createPayload(
+          PayloadType.PEER_JOINED,
+          '',
```
kanyesthaker, line 354: What is the second argument?

signaler/server.ts
```diff
@@ -197,71 +201,178 @@ export class SignalingServer {
...
+      }
+    });
+  }
+
+  /**
+   * Subscribe to the `MediaStream` of all peers in a room.
+   * @param {string} roomID ID of the room to get the peers from.
+   * @param {Peer} currentPeer The peer that wants to subscribe to all `MediaStream`s.
+   */
+  _subscribeToMediaStreams(roomID: string, currentPeer: Peer) {
```
kanyesthaker, line 368: What's an example for when we'd want to `_subscribeToPeerIDs` without also subscribing to the peer media streams? (i.e. why should these be separate)

signaler/utils/signal.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Look through the SDP messages and find the stream ID.
+ * @param {any} message The SDP message.
+ * @returns {string} The stream ID.
+ */
+export function checkICECandidateForStreamID(message: any) {
+  if (message) {
+    const sdp = message.sdp;
+    if (sdp) {
+      const streamID = sdp.match(/a=msid:(\w+)/);
```
kanyesthaker, line 10: please add a comment re: this line w/ the regex!

signaler/utils/signal.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Look through the SDP messages and find the stream ID.
+ * @param {any} message The SDP message.
+ * @returns {string} The stream ID.
+ */
+export function checkICECandidateForStreamID(message: any) {
```
kanyesthaker, line 6: nit: We're not actually checking the ICE candidate itself, right? rather, the SDP message, of which the candidate is a part?

signaler/server.ts
```diff
@@ -285,4 +397,132 @@ export class SignalingServer {
...
+    peer?.ws.send(signal);
+  }
+
+  /**
+   * Handle the `data` event from the SimplePeer listener.
+   * @param {any} data The data from the `data` event.
+   * @param {Peer} peer Peer that sent the `data` event.
+   */
+  _onPeerData(data: any, peer?: Peer) {
+    if (!peer) return;
```
kanyesthaker, line 457: Instead of manually checking `if (!peer) return;` for each of these functions, why not just make the `peer` a non-optional parameter? These should only be fired if a peer exists, right?

src/network/communicator.ts
```diff
@@ -16,20 +16,29 @@ export default class Communicator {
...
 
-  /** Returns whether or not we are connected to the signaling server. */
+  /** Returns whether or not we are connected to the SFU server via WS. */
   get isServerConnected(): boolean {
-    return this.messageRouter.isSocketConnected;
+    return this.messageRouter.isConnectedWS;
+  }
+
+  /** Returns the number of peers in the room. */
+  get numberOfPeersInRoom(): number {
```
kanyesthaker, line 35: nice convenience functions

src/network/hooks/useCommunicator.ts
```diff
@@ -25,35 +28,38 @@ const useCommunicator = ({
...
   const communicator = useAtomValue(communicatorAtom);
   const setIsPeerConnected = useSetAtom(isPeerConnectedAtom);
   const setIsServerConnected = useSetAtom(isServerConnectedAtom);
 
   // Set communicator callbacks
   useEffect(() => {
-    communicator?.on('connect', (message: any) => {
-      log.debug(`Connected to peer with message ${message}`);
+    communicator?.on('connect', (message, peerID) => {
+      log.debug(`Connected to peer ${peerID} with message ${message}`);
```
kanyesthaker, line 41: nit: why `log.debug` instead of `log.trace`? (since we're including the message body in the log message)

src/network/utils/signal.ts
```diff
@@ -0,0 +1,62 @@
+import log from 'loglevel';
+
+/**
+ * Look through the SDP messages and find the stream ID.
+ * @param {any} message The SDP message.
+ * @returns {string} The stream ID.
+ */
+export function checkICECandidateForStreamID(message: any) {
```
kanyesthaker, line 8: Is this function duplicated?

src/scene/players/Players.tsx
```diff
@@ -23,7 +24,16 @@ const Players = () => {
   };
 
   const onMessage = (message: any) => {
-    setPeerMovementData(JSON.parse(message));
+    if (message) {
```
kanyesthaker, line 27: nice error handling!

src/network/utils/payload.ts
```diff
@@ -111,24 +113,3 @@ export const createPayload = (
   };
   return JSON.stringify(payload);
 };
-
-/**
```
kanyesthaker, line 115: Nice cleanup

src/network/utils/cache.ts
```diff
@@ -0,0 +1,53 @@
+import log from 'loglevel';
+import RingBuffer from 'ringbufferjs';
+
+import { CommunicatorCallback, CommunicatorCallbackArgs } from './types';
+
+// Create buffers to store the data until the API user is ready to receive it.
+export function flush(
```
kanyesthaker, line 7: This file isn't super obvious to me as an outsider. Can we add a more descriptive docstring for the `flush` and `callbackHandler` functions that explain their role in the server?

src/network/messageRouter.ts
```diff
@@ -47,9 +49,46 @@ export default class MessageRouter {
 
   private onErrorCallback?: CommunicatorCallback;
 
-  isPeerConnected: boolean = false;
+  private onPeerJoinedCallback?: CommunicatorCallback;
 
-  isSocketConnected: boolean = false;
+  private onPeerLeftCallback?: CommunicatorCallback;
+
+  private peerErrorBuffer = new RingBuffer<CommunicatorCallbackArgs>(
```
kanyesthaker, line 56: Please add a comment describing the roles of these buffers.

Aside: for modularity, you could also make the object declaration of `RingBuffer<CommunicatorCallbackArgs>(MaxCachedMessages)` and wrap it in a class `Cache` and move it to `/cache.ts` -- would be a little simpler from an API design perspective. But this is probably overkill for how simple this is overall.

src/network/messageRouter.ts
```diff
@@ -361,8 +409,19 @@ export default class MessageRouter {
   /**
    * Handle a signal recevied from the other peer.
    * @param {any} signal The metadata containing the signal.
-   */
-  _handleServerSignal(signal: any) {
+   * @param {string} peerID The ID of the peer that sent the signal.
+   */
+  _handleServerSignal(signal: any, peerID: string) {
+    // Here we want to check if the signal we received contains a stream ID.
```
kanyesthaker, line 415: Nice solution to the problem of mapping Stream and Peer IDs together

src/network/messageRouter.ts
```diff
@@ -467,19 +540,52 @@ export default class MessageRouter {
...
+        log.debug(`Peer ${data.peerID} Joined Room`);
+        this.peersInRoom.push(data.peerID);
+        callbackHandler(
+          this.peerJoinedBuffer,
+          data.data,
+          data.peerID,
+          this.onPeerJoinedCallback,
+        );
+      } else if (data.type === PayloadType[PayloadType.PEER_LEFT]) {
+        log.debug(`Peer ${data.peerID} Joined Room`);
```
kanyesthaker, line 559:
```suggestion
        log.debug(`Peer ${data.peerID} Left Room`);
```

kanyesthaker (commented):
Really great work! I left some comments. Most of them are questions for my own understanding. Please take a look, after which I can come back and approve!

src/network/messageRouter.ts
shalins: Resolved in next PR :)

src/network/utils/cache.ts
shalins: Good call -- will update.

signaler/server.ts
shalins: Hmm, do we? This is a good catch, this might've broken STUN/TURN and I honestly have no idea how to test/implement this. Did you make a design doc or something that I can reference so I can maintain that support? I can also sync with you about this in person.

signaler/server.ts
shalins: Yes, this is "threadsafe" in the sense that the shared resource of `rooms` will not have an inconsistent value as a result of peers leaving in close succession before the loop has completed. Firstly, this is because NodeJS is single-threaded, and single threaded things are inherently thread-safe. Secondly, we have no concurrent operations, so the control flow and therefore order of instructions aren't interleaved. That being said, this is a good question, since there are still ways to get "race conditions" in a Node environment. One case is if you introduce any async code, since now you no longer can reason about what order the reads and writes happen, and instructions are interleaved. This most commonly happens when you make network requests and await a response, or you read from or write to a file on disk. Another case is if you truly have multiple CPU threads, which you can do through Node worker threads or using a different environment.

signaler/server.ts
shalins: This is just an empty string, since there's no data to send, we just want to send a message.

signaler/utils/signal.ts
shalins: Good point, will do!

src/network/hooks/useCommunicator.ts
shalins: So I think Sam added these debug logs to all of the callback functions originally, but I can remove the message from the log, I think it should still be debug since it would be confusing to see logs for some callbacks but not others.

src/network/utils/signal.ts
shalins: Yes, but not in the next PR.

signaler/server.ts
shalins: Yeah totally! For our use case, it's so light it doesn't really matter (we have really small N, since we probably won't have more than 20 peers in a given room)

signaler/server.ts
shalins: Good point, I will verify if there's an actual reason or not, and if not will just remove.

src/network/messageRouter.ts
shalins: Yeah good idea! Will update (maybe in the next PR since it's a larger change)

signaler/server.ts
shalins: Yeah I'll see if there was a reason I did this or not, if not will update!

signaler/utils/signal.ts
shalins: yup true, will change

signaler/server.ts
shalins: I think these functions each do two distinct things, so it makes sense to make them separate even if they're not used independently at the moment. Also its good practice since it makes things easier to reason about when I write unit tests.

signaler/server.ts
shalins: Was already fixed in the next PR

src/network/hooks/useCommunicator.ts
shalins: Update: This is just the `onConnect` handler, not the `onData`, so the message body should be trivially small, and might be helpful for debugging. Thus, I think it's fine to leave it.

signaler/server.ts
shalins: This was an awesome awesome catch btw @kanyesthaker! The `JSON.parse` was plain wrong and so the trace logs were not printing correctly, nice attention to detail!

signaler/server.ts
shalins: @kanyesthaker will need to discuss this with you actually. Keeping open for now.

signaler/server.ts
kanyesthaker: Maybe just add 1 extra sentence to the comment clarifying the intent behind this -- so just a quick sentence that says `[this is exclusively for local development]` or something? I'm OK with shipping this for now since it won't break production even if the package goes out of date.

src/network/messageRouter.ts
kanyesthaker: sounds good! in that case, can you add a `// TODO(Shalin)` here so we don't lose track?

signaler/server.ts
kanyesthaker: great, thanks for the context!

signaler/server.ts
kanyesthaker: awesome!!

signaler/server.ts
kanyesthaker: We can just do this observationally. I had written a private method that would intercept the ICE candidate callback in simplepeer and `log` the candidate information, and we can invoke that same function here. I can find a code pointer if helpful.

kanyesthaker: https://github.com/IrisHub/SpaceBar/blob/592bb8e2ea402ff8c4750a0b1296ac6c05ebad76/src/network/peerManager.ts#L97
and
https://github.com/IrisHub/SpaceBar/blob/592bb8e2ea402ff8c4750a0b1296ac6c05ebad76/src/network/peerManager.ts#L146

I'm not sure if there's a way to force seeing all the edge cases in the logs, but this method is sufficient for me

src/network/utils/cache.ts
shalins: Added better comments, will do a larger refactor of the cache in a future PR.

signaler/server.ts
shalins: Done

signaler/server.ts
shalins: So this is mainly protection against crashes due to opaque SimplePeer ordering. Since these functions are attached to SimplePeer handlers, I've experienced cases in which SimplePeer will fire onMessages even after I've called the "destroy" function, which sometimes causes a crash.

signaler/server.ts
shalins: Ok done, but I'm not sure how to interpret these, so lets talk about this when you're back.

kanyesthaker (approved):
:shipit:

PR #157, opened 2023-01-09
Switch from NPM to Yarn and upgrade critical packages

### LINKED ISSUE #149: Upgrade NPM packages to bring Three etc. to latest (stable) version

kanyesthaker:
### Description of Changes
- Installed yarn
- Upgraded critically outdated packages
- Fix all code paths that relied on now-deprecated or unsupported behavior

### Link to Issue
closes #149 

### Testing Procedure
TODO


package.json
```diff
@@ -25,24 +25,25 @@
...
+    "ws": "^8.8.0",
+    "yarn": "^1.22.19"
   },
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
     "format": "prettier --write \"**/*.{ts,tsx,json,md}\"",
     "test": "react-scripts test",
-    "eject": "react-scripts ejecti",
+    "eject": "react-scripts eject",
```
kanyesthaker, line 46: lol this has been misspelled forever

shalins: great catch lol

package.json
```diff
@@ -25,24 +25,25 @@
...
     "styled-components": "^5.3.5",
-    "three": "^0.138.0",
+    "three": "^0.148.0",
     "three-stdlib": "^2.8.11",
     "typescript": "^4.5.5",
     "uuid": "^9.0.0",
     "web-vitals": "^2.1.4",
-    "ws": "^8.8.0"
+    "ws": "^8.8.0",
+    "yarn": "^1.22.19"
```
shalins, line 39: just curious, why is yarn itself a package?

src/scene/models/Cylinder.tsx
```diff
@@ -24,7 +24,7 @@ const Cylinder = (props: CustomCylinder) => {
   let collisionRef = useRef<Object3D<Event>>(null);
 
   const meshProps = {
-    ref: props.collision ? collisionRef : undefined,
+    ref: props.collision ? collisionRef as Ref<Mesh<BufferGeometry>> : undefined,
```
shalins, line 27: Do you think it's a good idea to type alias this? Seems like we're using it all over the place, but idrc either way.

src/scene/models/PlayerModel.tsx
```diff
@@ -53,26 +59,29 @@ const PlayerModel = ({
...
-            <meshBasicMaterial attachArray="material" color={Colors.grey} />
-            <meshBasicMaterial attachArray="material" color={Colors.grey} />
-            <meshBasicMaterial attachArray="material" color={Colors.grey} />
-            <meshStandardMaterial attachArray="material" map={texture} />
+            <meshBasicMaterial attach="material-0" color={Colors.grey} />
+            <meshBasicMaterial attach="material-1" color={Colors.grey} />
+            <meshBasicMaterial attach="material-2" color={Colors.grey} />
+            <meshBasicMaterial attach="material-3" color={Colors.grey} />
+            <meshBasicMaterial attach="material-4" color={Colors.grey} />
+            <meshStandardMaterial attach="material-5" map={texture} />
```
shalins, line 84: Just curious, why are there numbers now?

src/scene/models/Sphere.tsx
```diff
@@ -25,7 +25,7 @@ const Sphere = (props: CustomSphere) => {
   let collisionRef = useRef<Object3D<Event>>(null);
 
   const meshProps = {
-    ref: props.collision ? collisionRef : undefined,
+    ref: props.collision ? collisionRef as Ref<Mesh<BufferGeometry>> : undefined,
```
shalins, line 28: On a second pass, I think this is a pretty ugly change that upgrading the 3js introduced, and seems like it touches every single model, since all of them have collision. Maybe it is better to type alias this since if anyone makes any new model, they'd have to remember to do this ugly typecasting thing when assigning props.

Also purely out of curiosity, what does `Ref<Mesh<BufferGeometry>>` mean vs what it was before?

src/scene/players/UserPlayer.tsx
```diff
@@ -22,7 +22,7 @@ interface UserPlayerProps {
  * according to keypress logic defined in ./playerMovementControls.ts.
  */
 const UserPlayer = ({ listener }: UserPlayerProps) => {
-  const PLAYER_VELOCITY_DECAY = 0.75;
+  const PLAYER_VELOCITY_DECAY = 0.2;
```
shalins, line 25: wonderful

shalins (approved):

src/scene/models/PlayerModel.tsx
kanyesthaker: good question, see this link https://github.com/pmndrs/react-three-fiber/discussions/744#discussioncomment-1521539

the "attachArray" argument was deprecated in favor of a universal "attach" property for the base Object3D class. The number helps specify the equivalent of an "index" in an array; so the argument accepts a more generic parameter, but the number helps tell Threejs that we're intending for this to be processed like an array.

src/scene/models/Cylinder.tsx
kanyesthaker: I think I prefer having this written explicitly, it's clearer as to what exactly this type is; plus this probably won't last long, the collision refactor might remove this entirely.

package.json
kanyesthaker: just how I installed it (via `npm`), equivalently you can install it via homebrew or similar. This just ensures that everyone's `yarn` stays sync'd

src/scene/models/Sphere.tsx
kanyesthaker: See my response above -- I worry that typecasting this will obfuscate what this object actually is, and I don't want us to write things like `collisionRef as SafeCollisionType` without fully understanding what it's doing.

Before, the `mesh` that accepts this `ref` would accept a generic `Ref<Object3D<Event>>`, but now Threejs has narrowed the type on the ref, so the ref for the `mesh` has to be a `Ref<Mesh<BufferGeometry>>`. But the hook that gives us the physics ref (from a different library) hasn't narrowed its type accordingly, so we have to cast down.

In the collision refactor we will likely be eliminating the usage of the offending hook entirely (details TBD). I'll keep this in mind and revisit it when I get there.

PR #164, opened 2023-01-22
feat: Migrate storage from S3 native to S3-compatible B2 with CloudFlare proxy

### LINKED ISSUE #163: Migrate all assets off of AWS S3

kanyesthaker:
### Description of Changes
- Migrate from AWS S3 to a BackBlaze B2 always-hot cloud storage bucket
- Set up a read-only proxy via CloudFlare for free egress
- Documentation and installation guide available at this [Notion page](https://www.notion.so/irisapp/Using-B2-Cloud-Storage-fac73aa9e3d64ffd9b30e2e2b2d2aa51) 
- Reformat "resolveStoragePath" function

In order to set up the CloudFlare proxy webworker, there is a new module `storage-api` with its own dependencies. The `worker.ts` file defines the actual webworker functionality by intercepting the `fetch` function, using the B2 S3-compatible API to pre-sign the request (read more [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ShareObjectPreSignedURL.html)), and then fetching the object. The `wrangler.toml` file defines metadata and environment variables referenced by the webworker. Private variables (such as the true API key) are directly maintained and encrypted in the CloudFlare console

### Link to Issue
closes #163 

### Testing Procedure
World loads properly!
image


package.json
```diff
@@ -20,6 +20,7 @@
     "@types/react-fontawesome": "^1.6.5",
     "@types/uuid": "^9.0.0",
     "@types/ws": "^8.5.3",
+    "aws4fetch": "^1.0.17",
```
samjgorman, line 23: What is this for? 

Read about the repo and its described as an AWS client. If we're no longer using AWS what role does this play?
https://github.com/mhart/aws4fetch

src/common/constants.ts
```diff
@@ -77,21 +77,21 @@ export abstract class Boundary {
...
-    resolveS3Path('calibration_skybox/positive-y.png', 'texture'),
-    resolveS3Path('calibration_skybox/negative-y.png', 'texture'),
-    resolveS3Path('calibration_skybox/positive-x.png', 'texture'),
-    resolveS3Path('calibration_skybox/negative-x.png', 'texture'),
+    resolveStoragePath('calibration_skybox/negative-z.png', Bucket.Texture),
+    resolveStoragePath('calibration_skybox/positive-z.png', Bucket.Texture),
+    resolveStoragePath('calibration_skybox/positive-y.png', Bucket.Texture),
+    resolveStoragePath('calibration_skybox/negative-y.png', Bucket.Texture),
+    resolveStoragePath('calibration_skybox/positive-x.png', Bucket.Texture),
+    resolveStoragePath('calibration_skybox/negative-x.png', Bucket.Texture),
```
samjgorman, line 94: Nice cleanup

src/common/utils/io.ts
```diff
@@ -1,21 +1,18 @@
+export enum Bucket {
+  Texture = 'textures/',
+  Model = 'models/',
+  Terrain = 'terrain/',
+}
+
 /**
  * Converts a shorthand file path to a full path for an S3 bucket.
```
samjgorman, line 8: Update docstring

storage-api/worker.ts
```diff
@@ -0,0 +1,110 @@
+// @ts-nocheck
+// Proxy Backblaze S3 compatible API requests, sending notifications to a webhook
```
samjgorman, line 2: I can use some help understanding why we need this.  Why are we proxying S3 compatible requests? For our migration?

storage-api/worker.ts
```diff
@@ -0,0 +1,110 @@
...
+  // Remove trailing slashes
+  path = path.replace(/\/$/, '');
+  // Split the path into segments
+  const pathSegments = path.split('/');
+
+  if (ALLOW_LIST_BUCKET !== 'true') {
+    // Don't allow list bucket requests
+    if (
+      (BUCKET_NAME === '$path' && pathSegments[0].length < 2) || // https://endpoint/bucket-name/
+      (BUCKET_NAME !== '$path' && path.length === 0)
```
samjgorman, line 44: Readability is taking a hit with these conditionals... Would recommend cleaning this up

samjgorman (commented):
Nice work moving quickly on this & taking initiative. Holding off on approving until I have a better understanding of what we're doing here, especially understanding the relationship of the Cloudflare web worker to Backblaze in order to avoid egress fees.  @kanyesthaker and I have already started to sync off Github & left a few relevant questions here

package.json
kanyesthaker: BackBlaze uses an S3-compatible API, which means that we can use the AWS S3 API in order to interact with BackBlaze directly. This AWS Client allows us to authenticate requests made by the CloudFlare worker.

src/common/utils/io.ts
kanyesthaker: good catch

storage-api/worker.ts
kanyesthaker: Answered async^

storage-api/worker.ts
kanyesthaker: good point -- this is code ripped from https://github.com/backblaze-b2-samples/cloudflare-b2 -- doesn't hurt to clean it up though

src/common/constants.ts
```diff
@@ -77,21 +77,21 @@ export abstract class Boundary {
...
-    resolveS3Path('posy.jpg', 'texture'),
-    resolveS3Path('negy.jpg', 'texture'),
-    resolveS3Path('posx.jpg', 'texture'),
-    resolveS3Path('negx.jpg', 'texture'),
+    resolveStoragePath('negz.jpg', Bucket.Texture),
+    resolveStoragePath('posz.jpg', Bucket.Texture),
+    resolveStoragePath('posy.jpg', Bucket.Texture),
+    resolveStoragePath('negy.jpg', Bucket.Texture),
+    resolveStoragePath('posx.jpg', Bucket.Texture),
+    resolveStoragePath('negx.jpg', Bucket.Texture),
```
shalins, line 85: Can `resolveStoragePath` not just take in a list to avoid copy-paste?

shalins (approved):
Approving, but just the small nitpick about overloading `resolveStoragePath` take in either a string or a list of strings to avoid code dup. Also, would you mind transpiling your Notion page into a README and placing in the storage-api dir? That would make it super easy to reference in the future I think. Apart from that, great work as always!

src/common/utils/io.ts
samjgorman: Doesn't look like this was resolved

storage-api/README.md
```diff
@@ -0,0 +1,3 @@
+# SpaceBar Storage API
```
samjgorman, line 1: Clear & concise!

samjgorman (approved):
LGTM! Thanks for this! Left one non-blocking nit

PR #165, opened 2023-02-07
Client Implementation for Multipeer Communicator

### LINKED ISSUE #28: Communicator should be able to connect more than 2 peers together over WS and WebRTC.

shalins:
### Description of Changes
This is the second part of #154, which enables the actual client support for multi-peer Communicator. This means 3+ clients will be able to join a room, see each other's video, hear each other's audio spatially, and be able to interact with each other and with the world.

### Link to Issue
Closes #28 

### Testing Procedure
Lots of testing with multiple people joining and leaving.
image


src/network/messageRouter.ts
```diff
@@ -548,18 +536,18 @@ export default class MessageRouter {
...
           data.data,
           data.peerID,
           this.onPeerJoinedCallback,
         );
       } else if (data.type === PayloadType[PayloadType.PEER_LEFT]) {
-        log.debug(`Peer ${data.peerID} Joined Room`);
-        this.peersInRoom = this.peersInRoom.filter(
-          (peerID) => peerID !== data.peerID,
-        );
+        log.debug(`Peer ${data.peerID} LEFT Room`);
```
shalins, line 547:
```suggestion
        log.debug(`Peer ${data.peerID} Left Room`);
```

sfu/server.ts
```diff
@@ -142,7 +145,6 @@ export class SignalingServer {
   _handleJoin(ws: WebSocket, roomID: string, peerID?: string) {
     log.debug(`Join Request Received with Room ID: ${roomID}`);
 
-    // Check if the room exists. If not, create it.
```
kanyesthaker, line 145: why are all these comments deleted?

sfu/utils/signal.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Look through the SDP messages and find the stream ID.
+ * @param {any} message The SDP message.
+ * @returns {string} The stream ID.
+ */
+export function checkICECandidateForStreamID(message: any): string | null {
+  if (message) {
+    const sdp = message.sdp;
+    if (sdp) {
+      const streamID = sdp.match(/a=msid:(\w+)/);
```
kanyesthaker, line 13: add comment explaining what's going on here

sfu/utils/signal.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }
+  return null;
+}
+
+// Overload the SimplePeer `send` method to check the `readyState` of the
+// RTCDataChannel before sending the message.
+export function peerSend(peer: SimplePeer.Instance, data: SimplePeerData) {
+  if (
+    (peer as any)?._channel &&
```
kanyesthaker, line 27: out of curiosity, what's going on with the typing here (`peer as any`)?

spacebar
```diff
@@ -110,7 +110,7 @@ case $main_param in
             if [[ "$deploy_arg" = "client" ]]; then
                 heroku builds:create -a spacebar
             elif [[ "$deploy_arg" = "server" ]]; then
-                cd signaler; heroku builds:create -a spacebar-5; cd ..
+                cd sfu; heroku builds:create -a spacebar-5; cd ..
```
kanyesthaker, line 113: nit: `sfu` is a pretty hard name to parse. Can we rename to `sfu_server` or just `server`? I think it's hard for a developer to understand this directory otherwise. Plus if we change the server configuration we have to rename the entire directory. We can mention that it's configured as an SFU server in the README.

src/interface/VideoPlayer.tsx
```diff
@@ -90,12 +91,13 @@ const VideoPlayer = ({ height, width }: VideoProps) => {
...
     try {
-      let stream = await navigator.mediaDevices.getUserMedia(permissions);
+      let stream = await navigator.mediaDevices?.getUserMedia(permissions);
       if (videoRef.current) {
-        videoRef.current.srcObject = new MediaStream(stream.getVideoTracks());
+        videoRef.current.srcObject = stream;
         videoRef.current.play();
       }
     } catch (err: any) {
+      log.error('Error starting media: ', err);
```
kanyesthaker, line 100: nice

src/network/hooks/useCommunicator.ts
```diff
@@ -77,26 +79,41 @@ const useCommunicator = ({
       if (onPeerJoined) {
         onPeerJoined(data, peerID);
       }
+      if (peerID) {
+        setOtherPeersInRoom((prev) => new Set(prev.add(peerID)));
```
kanyesthaker, line 83: Confused why there's different logic for joining/leaving. i.e.
`setOtherPeersInRoom((prev) => new Set(prev.add(peerID)));`
vs.
`otherPeersInRoom.delete(peerID); setOtherPeersInRoom(otherPeersInRoom);`

src/scene/players/PeerPlayer.tsx
```diff
@@ -26,59 +26,45 @@ import { PlayerModel } from '../models';
...
-      const inverseYQuat = new Quaternion(0, 1, 0, 0);
-      audioRef.current?.setRotationFromQuaternion(
-        audioRef.current?.quaternion.multiply(inverseYQuat),
-      );
-      audioRef.current?.setDirectionalCone(...AudioConstants.DirectionalCone); // Directional cone outside of which audio is muted
-
-      if (audioRef.current && process.env.REACT_APP_SANDBOX_FLAG === 'true') {
-        const helper = new PositionalAudioHelper(audioRef.current, 3, 10, 5);
-        audioRef.current.add(helper);
+    if (stream) {
```
kanyesthaker, line 43: nice cleanup!

src/scene/players/PeerPlayer.tsx
```diff
@@ -26,59 +26,45 @@ import { PlayerModel } from '../models';
...
+      if (stream.getAudioTracks().length > 0) {
+        const audioElement = document.createElement(
+          'audio',
+        ) as HTMLAudioElement;
+        audioElement.srcObject = stream;
+        audioElement.muted = true; // Only want to play from the PositionalAudio source
+        audioRef.current?.setMediaStreamSource(audioElement.srcObject);
+        // Handle spatial audio https://developer.mozilla.org/en-US/docs/Web/API/PannerNode
+        audioRef.current?.setRefDistance(AudioConstants.AudioRefDistance); // Distance at which audio decay begins
+        const rotationQuat = new Quaternion(...Object.values(quaternion));
```
kanyesthaker, line 62: Careful, from the diff it looks like this overwrites the PositionalAudio changes on the base branch

src/network/utils/signal.ts
```diff
@@ -60,3 +43,18 @@ export function setCustomIceCandidateCallback(peer: any) {
     }
   };
 }
+
+// Overload the SimplePeer `send` method to check the `readyState` of the
+// RTCDataChannel before sending the message.
+export function peerSend(peer: SimplePeer.Instance, data: SimplePeerData) {
+  if (
+    (peer as any)?._channel &&
```
kanyesthaker, line 51: ditto to previous comment re: typing

src/scene/players/Players.tsx
```diff
@@ -12,42 +14,60 @@ import UserPlayer from './UserPlayer';
...
-          videoStream={videoStream}
-          audioStream={audioStream}
-          listener={listenerRef}
-        />
+      {Array.from(otherPeersInRoom).map(
+        (peerId: string) =>
+          peerMovementData.get(peerId) && (
+            <PeerPlayer
+              playerMovementData={
+                peerMovementData.get(peerId) ?? {
```
kanyesthaker, line 61: Any shot that `otherPeersInRoom` and `peerMovementData` fall out of sync? For example, if there's a peer in `otherPeersInRoom` but no corresponding key in `peerMovementData`? In that case does it make sense to spawn the cube at `(0, 0, 0)`? Maybe instead of iterating through peerIDs in `otherPeersInRoom` we could iterate through `Object.keys(peerMovementData)` -- just a suggestion. If this is not a concern feel free to ignore.

src/scene/players/PeerPlayer.tsx
```diff
@@ -26,59 +26,45 @@ import { PlayerModel } from '../models';
 interface PeerProps {
   playerMovementData: PlayerMovementData;
   listener: MutableRefObject<AudioListener>;
-  videoStream?: MediaStream;
-  audioStream?: MediaStream;
+  stream?: MediaStream;
+  key: string;
```
kanyesthaker, line 30: Is this `key` prop actually used? I don't see it referenced in PeerPlayer

src/scene/players/Players.tsx
```diff
@@ -12,42 +14,60 @@ import UserPlayer from './UserPlayer';
...
+                  position: { x: 0, y: 0, z: 0 },
+                  quaternion: { x: 0, y: 0, z: 0, w: 0 },
+                }
+              }
+              stream={streamMap.get(peerId)}
+              listener={listenerRef}
+              key={peerId}
```
kanyesthaker, line 68: ditto to above, do we use this `key` prop?

src/scene/players/Players.tsx
```diff
@@ -12,42 +14,60 @@ import UserPlayer from './UserPlayer';
  */
 const Players = () => {
   // Define state variables and refs for managing players
-  const [peerMovementData, setPeerMovementData] =
-    useState<PlayerMovementData>(); // Manages player coordinates
-  const [videoStream, setVideoStream] = useState<MediaStream>(); // Manages player video stream
-  const [audioStream, setAudioStream] = useState<MediaStream>(); // Manages player audio stream
+  const [peerMovementData, setPeerMovementData] = useState<
```
kanyesthaker, line 17: These types (`Map<string, PlayerMovementData>`, `Map<string, MediaStream>`) are no longer trivial. Please update the comments for explainability!

kanyesthaker (commented):
great stuff as usual 🚀 left some comments and questions!

src/scene/players/PeerPlayer.tsx
shalins: It's required when we do a `forEach` so react knows the uniqueness of items, even if it's not used.

src/scene/players/Players.tsx
shalins: Answered in earlier comment.

src/scene/players/Players.tsx
shalins: There is already a `peerMovementData.get(peerId) &&` that prevents that case from happening, I have to put a default value here because Typescript forces me to. But yes, we should iterate through the peerID's in `otherPeersInRoom` because anything else will be inaccurate. In the case that they get out of sync due to latency, we should trust `otherPeersInRoom` not the other way around.

src/scene/players/Players.tsx
shalins: Sure, I'll update

src/scene/players/PeerPlayer.tsx
shalins: This change is intentional -- I made this change before PositionalAudio was fixed. I can make sure this doesn't break that though, thanks for pointing out.

sfu/server.ts
shalins: Not useful because obvious from looking at the code, or outdated.

src/network/hooks/useCommunicator.ts
shalins: `new Set(...)` is created because Set is pass by reference, which means if its elements get mutated, it doesn't trigger a refresh of a `useEffect` hook. Will make the delete consistent as well. Good question.

sfu/utils/signal.ts
shalins: Yup, will do

sfu/utils/signal.ts
shalins: Can't access `._channel` if peer is typed, since it's protected as private in the types .d.ts file. Will add a comment explaining.

spacebar
shalins: Good point, can rename to `server`.

src/network/utils/signal.ts
shalins: Responded to your earlier comment.

sfu/utils/signal.ts
shalins: Done!

sfu/utils/signal.ts
shalins: Done.

spacebar
shalins: Done

src/scene/players/Players.tsx
shalins: Done.

src/network/hooks/useCommunicator.ts
shalins: Done, seems to work fine

src/scene/players/PeerPlayer.tsx
shalins: Should be fixed, but will leave open so you can double check.

kanyesthaker (approved):
:shipit:

PR #169, opened 2023-02-15
fix: Correct the player positional audio orientation

### LINKED ISSUE #168: Bug: Positional audio plays out of the wrong side of the cube

kanyesthaker:
### Description of Changes
As part of the OpenGL Convention, the camera and its parent object face opposite directions (https://stackoverflow.com/questions/52957577/why-do-camera-and-object3d-look-opposite-directions). As a result, our positional audio would play out of the "back of the head" of the peer player instead of from their face.

This is rectified by multiplying the camera quaternion by (0, 1, 0, 0), representing a 180 degree rotation about the Y axis.

### Link to Issue
closes #168 

### Testing Procedure
Added a visualizer to see where the audio is actually playing from
image


src/scene/players/PeerPlayer.tsx
```diff
@@ -56,9 +63,18 @@ const PeerPlayer = ({
       audioRef.current?.setMediaStreamSource(audioElement.srcObject);
       // Handle spatial audio https://developer.mozilla.org/en-US/docs/Web/API/PannerNode
       audioRef.current?.setRefDistance(AudioConstants.AudioRefDistance); // Distance at which audio decay begins
+      // The camera faces the opposite direction from the object coords, so we multiply by (0, 1, 0, 0) to get it to face the right way
+      const inverseYQuat = new Quaternion(0, 1, 0, 0);
       const rotationQuat = new Quaternion(...Object.values(quaternion));
```
shalins, line 68: This is hard to reason about imo, what is the initial value of `rotationQuat`? Is this all being done in model space? If so, maybe explain that in the comment. 

Also are you sure having `0`'s in the inverseYQuat is a robust way to mirror an axis? What if rotationQuat initially starts with something like (0.2, 0.2, 0.2, 1), after the product, the new value will just be (0, 0.2, 0, 0) -- is that intended?

I'm sure you've thought through this -- I just wanted to get some additional clarity and double check so we avoid hard to catch bugs!

shalins (approved):

src/scene/players/PeerPlayer.tsx
kanyesthaker: All very good questions. `rotationQuat` is constructed from `quaternion` which is passed in from the player movement data. It is the camera's quaternion, represented in the local space of the camera. So "flipping" the camera to face the right way is equivalent to performing a rotation about the Y axis (yaw) in local object space.

Quaternion multiplication here is not element-wise. Multiplication of quaternions can be thought of as a composition (equivalent to matrix multiplication), where one rotation is performed and then the other; if I understand this correctly we are rotating the panner node to the correct location before rotating it 180 degrees about its (local) Y axis. I can add some comments to make this more clear.

This is a good comment though, and I'll make sure to double check that all of this is true before merging.

PR #170, opened 2023-02-16
feat: implement player-model collision

### LINKED ISSUE #56: Disable Jumping

### LINKED ISSUE #104: Peers should have a "personal space bubble"

### LINKED ISSUE #127: Player controls should feel smooth and natural

### LINKED ISSUE #186: Players should be able to collide with custom external meshes, and with other players

kanyesthaker:
### Description of Changes
Add player collision.
Player collision relies on the Bounding Volume Hierarchy implementation here https://github.com/gkjohnson/three-mesh-bvh. BVH trees are similar to other spatial data structures like Octrees and k-d trees.

At a high level, player collision works like this:
1. When we load a model, we aggregate all of its component meshes and geometries and create a BVH tree using the above package. This BVH tree is static (does not re-compute after initial load) and is stored in a state variable.
2. On each frame, the UserPlayer model checks whether it is intersecting a cell in the BVH tree. If it is, it will check all the mesh triangles inside that cell. It will then adjust the player's position so that it is no longer intersecting any meshes.
3. The same calculation happens for peer-peer collisions, except the BVH tree is not static, and is recomputed each frame.

### Link to Issue
closes #186 
closes #56 
closes #104 
closes #127 

### Testing Procedure

Outstanding bugs:
Players may get stuck inside one another: https://github.com/IrisHub/SpaceBar/issues/189
Players may clip through the floor: https://github.com/IrisHub/SpaceBar/issues/190
General performance improvements: https://github.com/IrisHub/SpaceBar/issues/191


src/common/utils/models.ts
```diff
@@ -0,0 +1,58 @@
...
+): Group => {
+  // Set model parameters
+  const model = gltf.scene;
+  model.position.copy(params.position);
+  model.scale.setScalar(params.scale);
+
+  // Traverse the glTF scene and extract each objects
+  model.traverse((obj) => {
+    if (obj instanceof Mesh) {
+      obj.castShadow = true;
```
samjgorman, line 29: Comment to describe why we need to set each object to `true`?

package.json
```diff
@@ -35,6 +36,7 @@
     "ringbufferjs": "^2.0.0",
     "styled-components": "^5.3.5",
     "three": "^0.148.0",
+    "three-mesh-bvh": "^0.5.21",
```
samjgorman, line 39: There's no types for `three-mesh-bvh`? From a quick search doesn't seem like it...

src/common/utils/models.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Generate a BVH tree (bounding box tree) and apply it to the geometry
+ * @param environment A group of Object3Ds to calculate the geometry for
+ * @returns The generated geometry with bounding box attributes
+ */
+export const applyBVHGeometry = (
+  environment: Group | Object3D[],
+): BufferGeometry => {
+  // Generate a static geometry from the environment
+  const staticGenerator = new StaticGeometryGenerator(environment);
+  staticGenerator.attributes = ['position'];
```
samjgorman, line 51: Would be nice if you could access an enum rather than using a string in this list here

src/common/utils/playerDynamics.ts
```diff
@@ -0,0 +1,168 @@
+import { Box3, Line3, Matrix4, Quaternion, Vector3 } from 'three';
+import { MeshBVH } from 'three-mesh-bvh';
+
+import { PhysicsConstants } from '../constants';
+
+type MovementControlsParams = {
+  left: boolean;
```
samjgorman, line 7: It's not obvious these values are booleans, as they're named closer to values. What if it was something like

````typescript
movedLeft: boolean
movedRight: boolean
movedForward: boolean
movedBackward: boolean


````

File: src/common/utils/playerDynamics.ts [hunk 5]
```diff
@@ -0,0 +1,168 @@
...
+  const playerDisplacement = new Vector3().subVectors(
+    adjustedPosition,
+    params.playerCurrentPosition,
+  );
+
+  return playerDisplacement;
+};
+
+/**
+ * A utility
```
samjgorman, line 144: What does this do? Document

src/scene/environment/Skybox.tsx
samjgorman, line 1: Would like to discuss this file during live sync

src/scene/models/GLTFCollisionModel.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+type GLTFCollisionModelProps = {
+  modelPath: string;
+  modelPosition: [number, number, number];
+  modelScale: number;
+};
+
+const GLTFCollisionModel = (props: GLTFCollisionModelProps) => {
+  /**
+   * The GLTFCollisionModel object downloads an object from storage,
```
samjgorman, line 24: Move to above function -- this isn't python! :)

samjgorman: Destructure the props you use rather than accessing the props object?

Ex) 

````
const GLTFCollisionModel = ({modelPath, modelPosition}: GLTFCollisionModelProps) => {
````

File: src/scene/models/GLTFCollisionModel.tsx [hunk 7]
```diff
@@ -0,0 +1,76 @@
...
+   * scales it appropriately, positions it within the world, and then
+   * uses utility functions to generate fine-grained collisions with the object
+   */
+
+  // Instantiage a Group, a collection of meshes
```
samjgorman, line 29: typo

src/scene/models/Plane.tsx
```diff
@@ -52,7 +52,7 @@ export default function Plane({ type = 'Static', ...props }: CustomPlane) {
         <meshStandardMaterial
           color={props.color}
           transparent={true}
-          opacity={props.opacity ? props.opacity : 1.0}
+          opacity={props.opacity ? props.opacity : 0.0}
```
samjgorman, line 55: Refactor to props.opacity ?? 0.0

src/scene/players/PeerPlayer.tsx
```diff
@@ -1,52 +1,90 @@
...
+  playerMovementDataMapAtom,
+  playerStreamMapAtom,
+} from '../../common/store';
+
+type Props = {
+  listener: AudioListener;
+  peerId: string;
+};
+
+const PeerPlayer = ({ listener, peerId }: Props) => {
```
samjgorman, line 33: Update component-level docstring

src/scene/players/PeerPlayer.tsx
```diff
@@ -1,52 +1,90 @@
...
+  const playerCollisionMesh = useRef<Mesh>(new Mesh());
+
+  // Generate the player collision geometry and its bounding box
+  playerCollisionMesh.current.geometry = new BoxGeometry(
+    ...PhysicsConstants.peerColliderBoundingBoxSize,
+  );
+  let staticGeometryGenerator = new StaticGeometryGenerator(
+    playerCollisionMesh.current,
+  );
+  staticGeometryGenerator.attributes = ['position'];
```
samjgorman, line 62: Why can't we use your `applyBVHGeometry`util?

src/scene/players/PeerPlayer.tsx
```diff
@@ -82,36 +120,56 @@ const PeerPlayer = ({ playerMovementData, stream, listener }: PeerProps) => {
...
-  };
-
-  return <PlayerModel {...videoPlayerModelProps} type={'Static'} />;
+  // On each frame, update the position of both the visual and collider mesh
+  useFrame(() => {
+    // Position update
+    if (movementData?.position) {
+      playerVisualMesh.current.position.lerp(
+        new Vector3(...Object.values(movementData.position)),
+        0.5,
```
samjgorman, line 129: What is 0.5 here? Refactor to constant?

src/scene/players/PeerPlayer.tsx
```diff
@@ -82,36 +120,56 @@ const PeerPlayer = ({ playerMovementData, stream, listener }: PeerProps) => {
...
+        1,
+      );
+    }
+  });
+
+  return (
+    <>
+      <mesh ref={playerVisualMesh}>
+        <boxBufferGeometry args={[1, 1, 1]} />
+        <>
```
samjgorman, line 155: Why do these meshes need to be wrapped in a fragment?

src/scene/players/PeerPlayers.tsx
```diff
@@ -0,0 +1,65 @@
...
+        );
+      } catch (e) {
+        log.error('Error parsing message', e);
+      }
+    } else {
+      log.error('Message was empty');
+    }
+  };
+
+  useCommunicator({ onMedia: onMedia, onMessage: onMessage });
```
samjgorman, line 47: Move any hook declarations to the top of the file for readability...

src/scene/players/PlayerCameraControls.tsx
```diff
@@ -19,5 +19,5 @@ export default function PlayerCameraControls() {
 
   // @ts-expect-error
   // Supressing known typing error with Drei: https://github.com/pmndrs/drei/issues/687
-  return <PointerLockControls enableDamping ref={controlsRef} />;
+  return <PointerLockControls pointerSpeed={0.3} ref={controlsRef} />;
```
samjgorman, line 22: Make this a constant

src/scene/players/UserPlayer.tsx
```diff
@@ -1,96 +1,217 @@
...
+  // The constants used here are just used for visualization/debugging
+  const playerColliderRef = useRef(
+    new Mesh(
+      new RoundedBoxGeometry(1, 2, 1, 5, PlayerConstants.playerCollisionRadius),
+      new MeshStandardMaterial(),
+    ),
+  );
+
+  // Instantiate player collision object
+  useEffect(() => {
```
samjgorman, line 91: Discuss during live sync

src/scene/players/UserPlayer.tsx
```diff
@@ -1,96 +1,217 @@
...
-      thisFrameVel.current.z = vel[2];
-    });
-  }, [playerPhysicsApi.velocity]);
 
   // Update movement vectors based on camera location
-  useFrame(() => {
+  useFrame((_, delta) => {
     // Copies the object's position to the camera (moves the camera)
-    playerPhysicsRef.current?.getWorldPosition(camera.position);
+    camera.position.lerp(playerColliderRef.current.position, 0.5);
```
samjgorman, line 125: Make this a constant?

src/scene/players/UserPlayer.tsx
```diff
@@ -143,7 +263,7 @@ const UserPlayer = ({ listener }: UserPlayerProps) => {
     }
   }, [communicator?.isConnectedP2P]);
 
-  return <mesh ref={playerPhysicsRef as Ref<Mesh<BufferGeometry>>} />;
+  return <mesh ref={playerColliderRef} />;
```
samjgorman, line 266: Nice! This is more type safe

yarn.lock
```diff
@@ -11293,7 +11293,7 @@ text-table@^0.2.0:
   resolved "https://registry.yarnpkg.com/text-table/-/text-table-0.2.0.tgz#7f5ee823ae805207c00af2df4a84ec3fcfa570b4"
   integrity sha512-N+8UisAXDGk8PFXP4HAzVR9nbfmVJ3zYLAWiTIoqC5v5isinhr+r5uaO8+7r3BMfuNIufIsA7RdpVgacC2cSpw==
 
-three-mesh-bvh@^0.5.23:
+three-mesh-bvh@^0.5.21, three-mesh-bvh@^0.5.23:
```
samjgorman, line 11296: `three-mesh-bvh `was already a dependency in our yarn.lock before you added it to the package.json?  Why is this? Same with `@types/three`?

samjgorman (commented):
Looking solid! Did a first pass, and looking forward to live review

.env
```diff
@@ -2,4 +2,4 @@
 REACT_APP_TERRAIN_FLAG=false
 
 # When true, render the 3-D calibration sandbox
-REACT_APP_SANDBOX_FLAG=false
\ No newline at end of file
+REACT_APP_SANDBOX_FLAG=false
```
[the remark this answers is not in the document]
kanyesthaker: I think I just added a newline to the end of this file, which I think we should have anyway

package.json
```diff
@@ -18,6 +18,7 @@
     "@types/react": "^18.0.26",
     "@types/react-dom": "^18.0.10",
     "@types/react-fontawesome": "^1.6.5",
+    "@types/three": "^0.148.0",
```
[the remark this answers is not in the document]
kanyesthaker: yeah, this was causing a ton of issues -- needed to add this in order to get player camera rotation speed to be editable

src/common/constants.ts
```diff
@@ -1,9 +1,15 @@
 import { Bucket, resolveStoragePath, resolveStoragePaths } from './utils/io';
 
 export abstract class PhysicsConstants {
-  static readonly gravity: [number, number, number] = [0, -30, 0];
+  static readonly gravity: number = -30;
 
   static readonly pointLight: [number, number, number] = [10, 10, 10];
+
+  static readonly collisionTickRate: number = 5;
```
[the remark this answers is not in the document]
kanyesthaker: yep got it. Since this is in a `common` folder at root instead of spatially close to the file, I think it's nice to have explicit type annotations so that a user in the `common` folder can read the file without needing to know where the constants are used; but it's weird to have this as a one-off case -- i can remove the type here or add type hints to the rest of the primitive constants

src/common/constants.ts
```diff
@@ -1,9 +1,15 @@
 import { Bucket, resolveStoragePath, resolveStoragePaths } from './utils/io';
 
 export abstract class PhysicsConstants {
-  static readonly gravity: [number, number, number] = [0, -30, 0];
+  static readonly gravity: number = -30;
```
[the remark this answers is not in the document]
kanyesthaker: ditto

src/common/store.ts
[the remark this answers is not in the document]
kanyesthaker: yeah I can do this, although all of these are really related to the `player` physics (the `player` is the only object in the world that actually has physics)

package.json
kanyesthaker: Don't think so

src/common/utils/models.ts
kanyesthaker: Hmmm, this would be an enum with only one entry, since `position` is the only attribute we could possibly want here -- it's the only attribute that will always be universally present in all glTF meshes, so it's the only one we can include

src/common/utils/playerDynamics.ts
kanyesthaker: good point, maybe `isMovingLeft` etc. -- I think this whole section can probably be cleaned up

src/common/utils/playerDynamics.ts
kanyesthaker: it is just a utility ¯\_(ツ)_/¯ lol yeah i guess I missed this comment

src/scene/environment/Skybox.tsx
kanyesthaker: yeah, tbh i think these changes should already be in main? i'm surprised they're not, i wonder if they were overwritten last time we merged in skybox. The TGA loader was meant to load a specific file format for the `debug` skybox

src/scene/models/GLTFCollisionModel.tsx
kanyesthaker: when is it right to do one over the other? I've seen both, wondering if there's a best practice

src/scene/models/GLTFCollisionModel.tsx
kanyesthaker: good eye! fyi you can make the change using <Cmd + g> while reviewing and commit the fix directly for small things like this :)

src/scene/players/PeerPlayer.tsx
kanyesthaker: hmmm I think I did this intentionally, but I don't remember why. I'll do some digging, I think we probably can use the util

src/scene/players/PeerPlayer.tsx
kanyesthaker: see the documentation for Vector3.lerp() -- I can make it a constant, but it's not something we're ever going to change, and it might hit readability if I have to reference something like `PlayerConstants.objectMovementLinearInterpolationConstant` -- maybe I can just make it a top-level constant within this file alone?

src/scene/players/PeerPlayer.tsx
kanyesthaker: React only allows returning a single JSX component in the return statement; multiple JSX components need to be returned in a fragment or similar; since we're returning multiple meshes here I've wrapped them in a fragment. LMK if there's a cleaner alternative!

src/scene/players/PeerPlayers.tsx
kanyesthaker: Hmmm, I'm not sure how I'd do this, because I only define `onMedia` and `onMessage` in this file

src/scene/players/UserPlayer.tsx
kanyesthaker: Could you be more specific about what you want to discuss?

src/scene/players/UserPlayer.tsx
kanyesthaker: Same idea as my previous `lerp()` comment^

yarn.lock
kanyesthaker: Yea not really sure what's goin on here lol

src/scene/players/UserPlayer.tsx
samjgorman: Just left these as markers to reference our live sync, not that helpful to enumerate the specifics async

src/scene/players/PeerPlayer.tsx
samjgorman: Yeah I think we can revisit our constants more holistically in a future PR.

TLDR:
If only used in one file, make a top level constant within the file
If used in a folder, have a folder-wide constants file that can be referenced
If reused across folders, add to our global constants file

src/scene/models/GLTFCollisionModel.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const environment = useRef(new THREE.Group());
+
+  // We store the mesh in Jotai to make it accessible across sibling components
+  // (i.e. components not in the same hierarchy in the DOM tree)
+  const [colliderMesh, setColliderMesh] = useAtom(environmentColliderAtom);
+  const collider = useRef(colliderMesh);
+
+  // Load the GLTF
+  new GLTFLoader().load(props.modelPath, (res) => {
```
kanyesthaker, line 38: Refactor this into a hook

src/scene/models/GLTFCollisionModel.tsx
```diff
@@ -0,0 +1,76 @@
...
+    // Set the collider mesh in state
+    setColliderMesh(collider.current);
+  });
+
+  // Bounding Volume Hierarchy visualizer tool
+  if (process.env.REACT_APP_SANDBOX_FLAG === 'true')
+    useHelper(collider, MeshBVHVisualizer);
+
+  // We have effectively decoupled the visual component of the object (environment)
+  // from the collision component (collider), so we return them separately
```
kanyesthaker, line 67:
```suggestion
  // from the invisible collision component (collider), so we return them separately
```

src/scene/players/PeerPlayers.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const otherPeersInRoom = useAtomValue(otherPeersInRoomAtom);
```
kanyesthaker, line 48: Lift this to the top of the file

src/common/utils/playerDynamics.ts
```diff
@@ -0,0 +1,168 @@
...
+ * Create a directional vector based on player keyboard inputs
+ * @param params A MovementControlsParams object that tracks player keyboard and mouse controls
+ * @returns A unit Vector3 that points in the appropriate direction in player-space
+ */
+export const getControlsMovementVector = (
+  params: MovementControlsParams,
+): Vector3 => {
+  // We only want the yaw with respect to global coordinates, so we normalize out the x and z quaternion components
+  // This allows us to move in the appropriate direction, even if we are looking straight up or straight down
+  const azimuthNormalization = Math.sqrt(
```
kanyesthaker, line 24: Change this to yaw

src/common/utils/playerDynamics.ts
```diff
@@ -0,0 +1,168 @@
...
+    params.angle.w ** 2 + params.angle.y ** 2,
+  );
+  // The azimuthal angle is our angle on the x-z plane (i.e. remove rotation on the y axis)
```
kanyesthaker, line 27: ditto

src/common/utils/playerDynamics.ts
```diff
@@ -0,0 +1,168 @@
...
+};
+
+/**
+ * A utility
+ * @param displacementVector
+ * @param currentTickGravity
+ * @param isPlayerOnGround
+ * @returns
+ */
+export const tickGravity = (
```
kanyesthaker, line 150: Rename to make more clear -- also add a comment

src/scene/players/UserPlayer.tsx
```diff
@@ -1,96 +1,217 @@
...
+          }),
+        );
+      }
+
+      // Adjust the player's position
+      playerColliderRef.current.position.add(collisionDisplacementVector);
+
+      // The player is on the ground if the player's collision adjustment is primarily
+      // "up" along the y-axis
+      isPlayerOnGround =
```
kanyesthaker, line 190: clean this up

kanyesthaker (commented):
- Some comments are obvious. We don't need those
- Some design decisions are very non-obvious. For example, why is the player a capsule shape? Can we move things like the GLTFLoader to a hook? Can we de-duplicate code (i.e. use useApplyBVH in the player collision file)? Are all points of complexity necessary (i.e. I'm probably using `lerp` incorrectly) 
- We should make sure these things are maintainable also. Make sure functions are named appropriately, constants are in the right place, and that everything is in a place where we can understand what's going on at high level.

src/common/utils/models.ts
```diff
@@ -43,11 +44,14 @@ export const retrieveGLTFMeshes = (
...
  */
-export const applyBVHGeometry = (
+export const generateBVHGeometry = (
   environment: Group | Object3D[],
 ): BufferGeometry => {
-  // Generate a static geometry from the environment
   const staticGenerator = new StaticGeometryGenerator(environment);
+
+  // The only property we care about for the BVH tree is position
+  // All others are unnecessary and can lead to issues with generation
```
samjgorman, line 53: This is helpful!

src/scene/players/UserPlayer.tsx
```diff
@@ -10,66 +10,68 @@ import { PhysicsConstants, PlayerConstants } from '../../common/constants';
...
 
 interface UserPlayerProps {
   listener: any;
 }
 
+/**
+ * This file renders the user player, calculates all player-based collisions, and
+ * handles streaming player movement and AV information to all peers. Information
+ * around collision detection can largely be understood by reading this blog post:
+ * https://wickedengine.net/2020/04/26/capsule-collision-detection/
```
samjgorman, line 30: This link may be better suited for a readme / documentation rather than a function level docstring

samjgorman (approved):
Let's get this in!  ⛽   Looks great -- thanks for going through this round of revs!

src/scene/players/UserPlayer.tsx
kanyesthaker: I think it's ok to keep this here, to keep the relevant information right next to the function where it's being implemented

PR #173, opened 2023-03-01
av: fixed bug with peer hearing their own audio

shalins:
### Description of Changes
Fixed the player hearing their own audio.

### Testing Procedure
Still need to test thoroughly with multiple peers, but so far it fixes the issue where I hear myself.


samjgorman (approved):
Thanks for the quick fix!

PR #179, opened 2023-03-06
video: added error when not connected to server

shalins:
### Description of Changes
Added error when the user is not connected to the signaling server, helpful for debugging issues.

### Link to Issue
Closes issue #178 

### Testing Procedure
Tested when the user has their video turned off, where it shows and error, and when there are multiple errors present. UI could be cleaned up, but out of scope for this PR.

image
image


src/interface/VideoPlayer.tsx
```diff
@@ -47,13 +50,40 @@ const VideoPlayer = ({ height, width }: VideoProps) => {
...
   };
 
+  /**
+   * Displays an error message to the user.
+   * @param err The error message to display.
+   */
+  const addError = (err: string) => {
+    setErrMessageOnUserMedia((prev) => {
+      if (!prev.includes(err)) {
+        return prev + ' ' + err;
```
samjgorman, line 68:
```suggestion
        return `${prev} ${err}`;
```

src/interface/VideoPlayer.tsx
```diff
@@ -47,13 +50,40 @@ const VideoPlayer = ({ height, width }: VideoProps) => {
...
+  };
+
+  /**
+   * Removes an error message from the user.
+   * @param err The error message to remove.
+   */
+  const removeError = (err: string) => {
+    setErrMessageOnUserMedia((prev) => {
+      if (prev.includes(err)) {
+        return prev.replace(err, '');
```
samjgorman, line 81: Why do we replace with an empty string rather than setting previous to be null or undefined?

src/interface/VideoPlayer.tsx
samjgorman: But why are we concatenating error messages together?

samjgorman: Doesn't need to be now: could be interesting to have a map that maps from error codes to whether they should be rendered to the end user. Then, the component responsible for rendering error messages maps over that map and renders a child error message component for each error message that should be rendered. So rather than concatenating error messages together, you can just manage setting the relevant messages' boolean state

samjgorman (approved):
Approving to unblock, left optional suggestion for a refactor as time permits

src/interface/VideoPlayer.tsx
shalins: Right now, the way error handling happens is done very naively, since even the AV errors don't go away if the user allows permissions (it requires a refresh). This is a larger refactor, so leaving out of scope for this PR.

Issue #175, opened 2023-03-03
Peer Audio does not work


kanyesthaker:

kanyesthaker:
As of 3/2/23 on the tip of main, the peer video will sometimes fail to render on the peer cube. Audio in the world seems to not work at all.

kanyesthaker:
Going to update this ticket to track audio fixes. As of 3/22, video works consistently, but Audio never does.

PR #181, opened 2023-03-12
communicator: promoted logs and fixed blank video bug

shalins:
### Description of Changes
* Promoted certain logs for easier debugging on the signaling server.
* Fixed the video not loading issue, root cause was messages receiving out of order, in which peer a received video from peer b before peer a knew the video was from peer b. Fixed by doing another check right before the callback is fired to see if the peer b's id came in yet. If we end up finding more synchronization issues even after this, it's worth thinking through a better model for syncing peer video and peer id.

### Link to Issue
Closes half of #175 (video not working part)

### Testing Procedure
Lots of testing with the world loaded on two different computers. 
IMG_2017


src/scene/players/PeerPlayer.tsx
```diff
@@ -76,7 +76,11 @@ const PeerPlayer = ({ playerMovementData, stream, listener }: PeerProps) => {
         }
       }
     }
-  }, [stream]);
+  }, [
+    stream,
```
kanyesthaker, line 80: nice!

src/network/utils/cache.ts
```diff
@@ -41,11 +46,28 @@ export function callbackHandler(
...
-    callback(...args);
-    log.info(`Flushing buffer: ${args} with ${buffer.size()} items left.`);
+    let [type, message, peerID] = args;
+
+    // Extra protection for the case where the peerID is not defined,
+    // which can happen if the media stream is received before the
+    // peer ID is received. In this case, do another lookup.
+    if (type === CallbackType.MEDIA) {
+      if (!peerID && peerIDMap) {
+        peerID = peerIDMap.get((message as MediaStream)?.id);
```
kanyesthaker, line 60: 🚀

kanyesthaker (approved):
nice! Appreciate the careful and intentional logging

src/network/messageRouter.ts
```diff
@@ -604,15 +608,25 @@ export default class MessageRouter {
...
+        this.peerMediaBuffer,
+        stream,
+        this.streamIDToPeerIDMap.get(stream.id),
+        this.onMediaCallback,
+      );
+    } else {
+      log.info(
+        `Couldn't find peerId for stream ${stream.id}. Retrying in ${StreamPeerCheckInterval}}ms`,
+      );
+      setTimeout(() => {
```
kanyesthaker, line 626: nit: normally when applications have a timeout-based retry interval, I've seen the timeout limit grow -- so the first time 100ms, then 200ms, then 500ms, etc. to make sure we never get in a state where we're doing infinite retries on a 100ms interval. Don't think it's important now, but a gentle suggestion

kanyesthaker (approved):

src/network/messageRouter.ts
shalins: Sure, its called backoff. We can do that here

shalins: Done

PR #193, opened 2023-03-26
fix: player will no longer get stuck in floor as the world loads

### LINKED ISSUE #190: Players may get stuck in the floor on-spawn

kanyesthaker:
### Description of Changes
Fix the "player gets stuck in floor" bug by fixing the player y-position to be above floor level, and also fixing a bug in how player height was set (player height previously failed to take into account the "radius" of the capsule).

### Link to Issue
closes #190 

### Testing Procedure
Tested by repeated reloading the world, and trying to see if the "getting stuck" happened with multiple peers.


src/scene/players/UserPlayer.tsx
```diff
@@ -107,7 +107,13 @@ const UserPlayer = ({ listener }: UserPlayerProps) => {
       // This has to be a Line3 due to how BVH is implemented, but it is conceptually a ray
       ray: new Line3(
         new Vector3(0, 0, 0),
-        new Vector3(0, -PlayerConstants.playerHeight, 0),
+        new Vector3(
+          0,
+          -(
+            PlayerConstants.playerHeight - PlayerConstants.playerCollisionRadius
```
samjgorman, line 113: Can these variables ever get out of sync with each other?  For example, what happens if I dramatically widen the `playerCollisionRadius`?  Slight concern that modifying these constants may have non-obvious effects here, but may not be following why you need to do this...

samjgorman (commented):
Nice work! Had a quick question, otherwise LGTM!

src/scene/players/UserPlayer.tsx
kanyesthaker: image
This diagram should help explain what's going on. The only time you're in trouble is if you make the collision radius larger than the height of the player. I think it's appropriate that arbitrarily changing constants will have consequences in behavior; I have left a comment in constants.ts warning of this though

samjgorman (approved):
Awesome!

PR #197, opened 2023-04-01
feat: Update player model

### LINKED ISSUE #189: Players will sometimes get stuck inside one another

kanyesthaker:
### Description of Changes
This pull request updates the player model to use a custom-optimized glTF, initially adapted from SketchFab. The purpose of this is twofold:

1. I hypothesize that making the player model smaller and friendlier will make the space feel more comfortable and reduce the friction with using the mouse to move (i.e. if a player is a cube, rotating with the mouse feels like the whole body is rotating. But if the player is a sphere, maybe it will feel more like only the player's head is rotating, which feels more natural).
2. Changing the hitbox of the player to only exist where the camera is (not on all sides) preserves the "personal space bubble" feeling without the annoyance of having a large impassible object in the bar.
3. Shrinking the player collision bounding box solves the problem of multiple peers getting stuck inside one another. This may still happen occasionally (less than 1 in 20 times), but moving the mouse should fix the problem.

Committed the dvc hash of the blender character model, which was optimized for performance for a size of roughly 200kb + added accent elements to the base design

### Link to Issue
closes #189 

### Testing Procedure
Tested locally. Plan on doing joint testing session with Sam before merging.


src/common/constants.ts
```diff
@@ -96,7 +96,7 @@ export abstract class SkyboxImages {
 }
 
 export abstract class PlayerConstants {
-  static readonly playerSize: [number, number, number] = [0.5, 0.5, 0.5];
+  static readonly playerSize: [number, number, number] = [1, 1, 0.125];
```
samjgorman, line 99: Nice bump in size -- noticed the improvement in live QA

src/common/utils/models.ts
```diff
@@ -21,6 +29,7 @@ export const getProcessedGLTFMeshes = (
   const model = gltf.scene;
   model.position.copy(params.position);
   model.scale.setScalar(params.scale);
+  model.quaternion.copy(params.quaternion);
```
samjgorman, line 32: Why do we do this?

src/scene/players/PeerPlayer.tsx
```diff
@@ -1,22 +1,23 @@
+/* eslint-disable react/display-name */
```
samjgorman, line 1: Can we avoid needing to disable this lint rule?

src/scene/players/PeerPlayer.tsx
```diff
@@ -126,45 +148,87 @@ const PeerPlayer = ({ listener, peerId }: Props) => {
...
+
+      playerCollisionMesh.current.position.copy(newPosition);
 
       // Refit the collider boundsTree
       playerCollisionMesh.current.geometry.boundsTree?.refit();
+    } else {
+      model.current.position.add(offset);
     }
+    t.current = t.current + delta;
+    console.log(`T: ${t.current}`);
```
samjgorman, line 170: 👀 Let's remove these logs!

src/scene/players/PeerPlayer.tsx
```diff
@@ -126,45 +148,87 @@ const PeerPlayer = ({ listener, peerId }: Props) => {
   }, [peerStream, audioTracksLength, videoTracksLength]);
 
   // On each frame, update the position of both the visual and collider mesh
-  useFrame(() => {
+  const t = useRef<number>(0);
```
samjgorman, line 151: Need a more descriptive variable name here -- does this represent time?

src/scene/players/PeerPlayer.tsx
```diff
@@ -126,45 +148,87 @@ const PeerPlayer = ({ listener, peerId }: Props) => {
...
+  useFrame((_, delta) => {
     // Position update
```
samjgorman, line 153: This comment isn't descriptive enough to be helpful -- remove or add a more descriptive comment that talks about how this adds a slight bobbing to the player model

src/scene/players/PeerPlayer.tsx
```diff
@@ -83,6 +92,19 @@ const PeerPlayer = ({ listener, peerId }: Props) => {
     );
   }, []);
 
+  const model = useRef(new Group());
+  useEffect(() => {
```
samjgorman, line 96: Let's add a comment to this useEffect to describe what we're doing here

samjgorman (changes requested):
Awesome work here, left comments to see this get cleaned up a bit more and we'll be ready to merge!

src/common/utils/models.ts
kanyesthaker: In case we want to be able to rotate the model

src/scene/players/PeerPlayer.tsx
samjgorman: Don't think this was addressed?

samjgorman (approved):
Looks good, looks like one point wasn't addressed but otherwise LGTM!

PR #201, opened 2023-04-03
server: server audio & 2d self video component fixes

### LINKED ISSUE #175: Peer Audio does not work

shalins:
### Description of Changes
* Fixes the issue in which peers were not able to hear each other's audio, in spite of the stream being received on both ends.
* Fixes the issue in which sometimes, the 2D component does not properly render the webcam video on the top left corner.

### Link to Issue
Fixes #175 and #200

### Testing Procedure
Tried with multiple peers on a single laptop, as well as many tests with an echo server, and was able to hear audio. Though there were things broken with positional audio that prevented the sound from being played correctly. Also saw that the video feed on the top left was back consistently.


kanyesthaker (approved):
awesome 🚀

PR #205, opened 2023-04-09
GLTF without collisions + seatmap

samjgorman:
### Description of Changes

- Support for a new `seatMapPath`
- `GLTFModel` prop to toggle collisions 
- Use this `GLTFModel` to render the `seatMap` GLTF in `SceneEntities`
- Committed blender file with scaled width world + grouped seatmaps + bar to dvc

### Link to Issue
https://github.com/IrisHub/SpaceBar/issues/204

### Testing Procedure

image

- Live QA session with Kanyes


src/scene/models/GLTFModel.tsx
```diff
@@ -0,0 +1,33 @@
+import { BoxProps } from '@react-three/cannon';
```
kanyesthaker, line 1: This file seems like it was revived from an earlier iteration of SpaceBar. can we try to refactor/repurpose the existing GLTFCollisionModel component so that we can toggle on/off collision? I'm hesitant to re-introduce this, because it (a) uses packages like `@react-three/cannon` which we do not use anymore, and (b) is less interpretable/readable than the current structure of our component library

kanyesthaker (commented):
Thanks for adding this! I think we should probably refactor the existing glTF loading logic to accommodate non-collision objects instead of reviving the old glTFModel component, which uses paradigms/dependencies which we no longer support.

src/scene/entities/SceneEntities.tsx
```diff
@@ -1,15 +1,24 @@
...
+import GLTFModel from '../models/GLTFModel';
 
 const SceneEntities = () => {
   return (
-    <GLTFCollisionModel
-      modelPath={Objects.barPath}
-      modelPosition={[0, 0, 0]}
-      modelScale={1}
-    />
+    <>
```
kanyesthaker, line 8: Can we add a comment describing what the "Seat Map" is, and why we need it? Will help with future maintainability.

src/scene/models/GLTFModel.tsx
```diff
@@ -13,22 +13,24 @@ import {
...
  * scales it appropriately, positions it within the world, and then
  * uses utility functions to generate fine-grained collisions with the object
  */
-const GLTFCollisionModel = ({
+const GLTFModel = ({
   modelPath,
   modelPosition,
   modelScale,
-}: GLTFCollisionModelProps) => {
+  applyCollisions = true,
```
kanyesthaker, line 32: Since generating collision is an expensive operation, perhaps we should make the default false? what do you think?

kanyesthaker (approved):
great! a few minor comments, but approving. Thanks for the addition!

src/scene/entities/SceneEntities.tsx
samjgorman: Done!

src/scene/models/GLTFModel.tsx
samjgorman: Default false is reasonable, switched!

PR #208, opened 2023-04-20
fix: fix buggy state and refactor player components

### LINKED ISSUE #209: Peer audio and video are either laggy and distorted or do not work at all

kanyesthaker:
### Description of Changes
This PR fixes broken peer audio/video by adjusting the way we grab media from state. Additionally, it refactors player-related components to isolate highly-rerendering components away from those which should not rerender.

### Link to Issue
Closes #209 

### Testing Procedure
Tested with Sam in 2 QA sessions, with 3 laptops.


samjgorman (commented):
It's looking solid!

My main question is whether we can separate out mediaStream data fetching logic with render logic.  I put up a draft PR to illustrate a potential approach where we define two new hooks related to fetching the user and player's media streams. Let's discuss offline and converge on a solution  
https://github.com/IrisHub/SpaceBar/pull/210/files

samjgorman (approved):
This is big!!! Excited for this

PR #219, opened 2023-04-30
feat: clientside caching for models and textures

### LINKED ISSUE #218: Textures (skybox) will flicker as the world loads

kanyesthaker:
### Description of Changes
The React Fiber (Drei) `useLoader` (and the derived `useCubeTexture`) hooks caches textures and models. This prevents any unnecessary re-downloads of assets, and should solve the problem where the skybox "flickers" as the rest of the world is loading.

### Link to Issue
closes #218 

### Testing Procedure
Tested locally. No functional differences, but all textures and objects now only load once

Before this PR:
image

After this PR:
image


src/scene/models/GLTFModel.tsx
```diff
@@ -40,34 +38,34 @@ const GLTFModel = ({
...
+      quaternion: new Quaternion(0, 0, 0, 1),
+    });
 
-      environment.current.add(modelEnv);
+    environment.current.add(modelEnv);
 
-      if (applyCollisions) {
-        // Generate the bounding box
-        collider.current.geometry = generateBVHGeometry(environment.current);
+    if (applyCollisions) {
```
samjgorman, line 53: Does `applyCollisions` need to be a dependency in this useEffect?

samjgorman (commented):
Have a question about whether we're capturing the full set of dependencies in the specified useEffect, otherwise looks good

src/scene/models/GLTFModel.tsx
kanyesthaker: applyCollisions is a prop right? it won't change until the whole component is rerendered, so I don't think we need to have it as a dep?

samjgorman: If you omit these props from the dependency array, the effect will only run once when the component mounts OR changes to gltf and don't think will react to changes in props. Since a change in props triggers a component update, not a remount, thinking this could open us up to subtle bugs where the value of `applyCollisions` changes, then `gltf` changes, the useEffect is re-run, and then an incorrect code path is used.

So it seems like we should add add `applyCollisions`, `modelPath`, `modelPosition` and `modelScale` as dependencies in this useEffect

kanyesthaker: ah got it. I don't think there's any circumstance where the `applyCollisions` value would change once it's initially set -- it's pretty much a static property of the object being rendered. The same is true for the modelPath, modelPosition, and modelScale -- these are initialization parameters that will not be changed. However, I don't see any harm in including them in the deps array, since it's just an extra safety check.

samjgorman (approved):

Issue #220, opened 2023-05-01
Better loading UX 


samjgorman:
Users are currently placed into a world that is actively loading GLTF models and can get themselves into a set of dangerous states, such as running out of the boundaries of the world before the full GLTF loads.

PR #221, opened 2023-05-01
Improve initial loading with loading screen + safer key controls

samjgorman:
### Description of Changes

- Only add event listeners to the window upon GLTFs loading into the world on load to prevent user from moving around the world before it loads and getting into a dangerous state
- Add an off-the-shelf `<Loader/>` component from `Drei` that provides a simple loading UI while GLTF models are initially loading. We can and should customize this, but this provides a solid foundation to build upon. 

https://user-images.githubusercontent.com/31233283/235414223-ec6dedf8-04a7-4c3d-a4f0-868ae55e39fb.mp4

### Link to Issue
https://github.com/IrisHub/SpaceBar/issues/220

### Testing Procedure
Locally tested for two peers across multiple refreshes, verified that new user joining the space does not block player movement for the user


src/scene/models/GLTFModel.tsx
```diff
@@ -46,7 +46,7 @@ const GLTFModel = ({
       const modelEnv = getProcessedGLTFMeshes(res, {
         position: new Vector3(...modelPosition),
         scale: modelScale,
-        quaternion: new Quaternion(0, 0, 0, 1)
+        quaternion: new Quaternion(0, 0, 0, 1),
```
kanyesthaker, line 49: Check your ESLint rules -- should we have comma dangle on or off?

src/scene/players/UserPlayerEntity.tsx
```diff
@@ -49,47 +51,46 @@ const UserPlayerEntity = ({ listener }: UserPlayerEntityProps) => {
   let isPlayerOnGround = false;
   let playerGravity = new Vector3();
 
-  window.addEventListener('keydown', (e) => {
+  const isUserPlayerControlsEnabled = modelLoadingProgress === 100;
+
+  const keyDownHandler = (e: KeyboardEvent) => {
     switch (e.code) {
       case 'KeyW':
-      case 'ArrowUp':
```
kanyesthaker, line 55: Why are we removing arrow keys?

src/scene/players/UserPlayerEntity.tsx
```diff
@@ -49,47 +51,46 @@ const UserPlayerEntity = ({ listener }: UserPlayerEntityProps) => {
   let isPlayerOnGround = false;
   let playerGravity = new Vector3();
 
-  window.addEventListener('keydown', (e) => {
+  const isUserPlayerControlsEnabled = modelLoadingProgress === 100;
```
kanyesthaker, line 54: Let's use a more generic name than `modelLoadingProgress` since ideally this isn't just the model -- all assets (including skybox, etc.) should be captured in the loader. Even if this isn't how the loader functions now, it's how it should work in the future.

src/scene/Scene.tsx
```diff
@@ -36,13 +36,20 @@ const Scene = () => {
...
           </>
         )}
+
         <PointerLockControls pointerSpeed={PlayerConstants.pointerSpeed} />
         <Skybox />
-        <SceneEntities />
+        {/** Fallback is null for suspense because we use <Loader/> to
+         * render a loading state instead
+         */}
+        <Suspense fallback={null}>
```
kanyesthaker, line 45: Can't we just make the fallback for the `<Suspense/>` be `<Loader/>`? This is based on the react fiber docs https://docs.pmnd.rs/react-three-fiber/tutorials/loading-models#showing-a-loader

src/scene/Scene.tsx
```diff
@@ -36,13 +36,20 @@ const Scene = () => {
...
+          <SceneEntities />
```
kanyesthaker, line 46: We should put the Skybox inside the Suspense as well, no?

kanyesthaker (commented):
great QoL feature 🚀 added some comments

src/scene/players/UserPlayerEntity.tsx
samjgorman: Hmm.. this shouldn't be happening. Will bring back!

src/scene/Scene.tsx
samjgorman: We actually can't do this, the `loader` won't load. believe the Drei specific `Loader` needs to be rendered outside of the canvas, for example, as it's used in the docs here
https://github.com/pmndrs/drei#loader

Would rather follow this approach if we were implementing a custom loader

src/scene/Scene.tsx
samjgorman: Sure!

src/scene/players/UserPlayerEntity.tsx
samjgorman: Updated to `assetsLoadingProgress`

src/scene/models/GLTFModel.tsx
samjgorman: Discussed off github & decided to keep comma dangling on

This comes from mismatch between prettier and eslint. Prettier enables by default while ESLint does not.

It's possible to override prettier by supplying a prettier config in our `spacebar`file, but given that multiple files already have comma-dangle, decided to not go down this route

PR #226, opened 2023-06-01
Enable sentry for FE monitoring

samjgorman:
### Description of Changes

- Enable Sentry to log and track FE errors

Example: pointer lock error getting logged in sentry
image

Will write a brief notion doc describing how to access sentry and how to get value from it

### Link to Issue
https://github.com/IrisHub/SpaceBar/issues/227

### Testing Procedure
Created and logged a dummy error


src/index.tsx
```diff
@@ -14,8 +15,41 @@ if (process.env.REACT_APP_LOG_LEVEL) {
   log.setLevel(process.env.REACT_APP_LOG_LEVEL as log.LogLevelDesc);
 }
 
+// Init sentry logging
+Sentry.init({
+  dsn: 'https://aab1fccd8f4649f69c22b5721ad767ce@o4505282763161600.ingest.sentry.io/4505282764800000',
```
kanyesthaker, line 20: are we leaking a token here?

src/index.tsx
```diff
@@ -14,8 +15,41 @@ if (process.env.REACT_APP_LOG_LEVEL) {
...
+  integrations: [new Sentry.BrowserTracing({}), new Sentry.Replay()],
+  // Performance Monitoring
+  tracesSampleRate: 1.0, // Capture 100% of the transactions, reduce in production!
```
kanyesthaker, line 23: what does "reduce in production" mean? how are we doing it?

kanyesthaker (commented):
thanks for adding this! make sure to link a github issue (create one if it doesn't exist), I do want to build that habit. Could you also link your notion doc in the PR description?

src/index.tsx
samjgorman: https://docs.sentry.io/product/sentry-basics/dsn-explainer/

Great question! This is a datasource name (DSN) that Sentry specifies is safe to have publically available

src/index.tsx
samjgorman: `tracesSampleRate determines the representative sample of errors sent to Sentry.

From the sentry docs: 

We recommend sampling your transactions for two reasons:

1. Capturing a single trace involves minimal overhead, but capturing traces for every page load or every API request may add an undesirable load to your system.
2. Enabling sampling allows you to better manage the number of events sent to Sentry, so you can tailor your volume to your organization's needs.

---

Because our system isn't particularly large and have not yet seen noisey errors crowding out information we need, would recommend leaving the sampling rate at 100%.  Let's observe the effect this has on Spacebar (from initial QA did not notice any performance regressions from leaving this on) and reduce if necessary from our next QA?

kanyesthaker: sgtm

kanyesthaker (approved):

PR #235, opened 2023-06-21
Fix: Show player's own video with better error handling

samjgorman:
### Description of Changes

- Fix bug where player's own video would not show. This is because the component would switch between rendering the error message dialog component to the video component and lose the media stream context
- This "You are not connected to the server" message isn't part of the standard errors that media stream gives, and isn't actionable + is too technical for an end user -- so lifted this down to the log level where it can be accessed in the info console. 

### Link to Issue
https://github.com/IrisHub/SpaceBar/issues/200

### Testing Procedure
Tested locally that player can see own video on join + when peer joins


src/interface/VideoPlayer.tsx
```diff
@@ -51,7 +53,9 @@ const VideoPlayer = ({ height, width }: VideoProps) => {
   const [audioEnabled, setAudioEnabled] = useState(true);
   const [videoEnabled, setVideoEnabled] = useState(true);
   const [errMessageList, setErrMessageList] = useState<string[]>([]);
-  const [errMessage, setErrMessage] = useState<string | undefined>(undefined);
+  const [showErrMessage, setShowErrMessage] = useState<string | undefined>(
```
kanyesthaker, line 56: Why the name change? The thing that we're setting in this variable is the actual error message text itself, right?

src/interface/VideoPlayer.tsx
```diff
@@ -222,9 +213,9 @@ const VideoPlayer = ({ height, width }: VideoProps) => {
 
   return (
     <>
-      {errMessage && <Dialog text={errMessage} color={Colors.warningRed} />}
-
-      {!errMessage && (
+      {showErrMessage ? (
+        <Dialog text={showErrMessage} color={Colors.warningRed} />
+      ) : (
```
kanyesthaker, line 218: If we're moving the AVNotConnectedInfoMessage to the log, do we still need to conditionally render the video for the remaining error message types? Notice you removed the `!errMessage` conditional

kanyesthaker (commented):
looks p good, just a couple qs

src/interface/VideoPlayer.tsx
samjgorman: Good catch, changed this back

src/interface/VideoPlayer.tsx
kanyesthaker: sgtm!

PR #236, opened 2023-06-28
fix: fix improper memory management in rendering code

### LINKED ISSUE #232: Client-side memory leak with multiple peers

kanyesthaker:
### Description of Changes
Refactor all useFrame loops to not allocate any new memory. There should be no allocation (`const`, `let`, `new`) of non-primitive objects within these loops (allocating primitives is OK).

Additionally makes player movement a Ref (to avoid unnecessary rerenders) and makes player video stateful (to trigger necessary rerenders).

### Link to Issue
closes #232 

### Testing Procedure
Pure code change, no functional change expected


src/common/utils/playerDynamics.ts
```diff
@@ -11,42 +11,51 @@ type MovementControlsParams = {
   angle: Quaternion;
 };
 
+const GlobalControlsObjects = {
```
samjgorman, line 14:
```suggestion
type GlobalControlsObjectsProps = {
...
}
const GlobalControlsObjects: GlobalControlsObjectsProps = {
```

src/common/utils/playerDynamics.ts
```diff
@@ -57,32 +66,47 @@ type PlayerCollisionParams = {
   boundsTree?: MeshBVH;
 };
 
+const GlobalCollisionsObjects = {
```
samjgorman, line 69:
```suggestion
type GlobalCollisionsObjectsProps = {
...
}

const GlobalCollisionsObjects: GlobalCollisionsObjectsProps  = {
```

src/common/utils/playerDynamics.ts
```diff
@@ -57,32 +66,47 @@ type PlayerCollisionParams = {
...
-  let collisionRay = new Line3().copy(params.playerRay);
+  collisionRay.copy(params.playerRay);
 
   collisionRay.applyMatrix4(params.playerWorldMatrix);
 
   // Create a box representing the bounding box of the player
   // The ray is the central "axis" of the player capsule, and we then expand
   // the bounding box to account for the "radius"
-  let playerCollisionBoundingBox = new Box3();
+  playerCollisionBoundingBox.makeEmpty();
```
samjgorman, line 104: Why are we making this bbox empty?

src/common/utils/playerDynamics.ts
```diff
@@ -152,17 +169,18 @@ export const getUpdatedGravityVector = (
...
-    return new Vector3().addVectors(
+    target.addVectors(
       currentTickGravity,
       displacementVector.multiplyScalar(
         -displacementVector.dot(currentTickGravity),
       ),
     );
   } else {
-    return new Vector3();
+    target.fromArray([0, 0, 0]);
```
samjgorman, line 184: Should [0,0,0] be a constant we store instead?  And should the `y` be slightly higher than 0 to prevent peers from being stuck in the floor?

src/scene/players/PeerPlayerEntity.tsx
```diff
@@ -14,18 +13,17 @@ import {
...
   generateBVHGeometry,
   getProcessedGLTFMeshes,
 } from '../../common/utils/models';
 
 type PeerPlayerEntityProps = {
   peerId: string;
-  videoRef: RefObject<VideoTexture | undefined>;
+  playerDataMapRef: MutableRefObject<Map<string, PlayerMovementData>>;
+  playerCollisionMapRef: MutableRefObject<Map<string, Mesh>>;
+  videoTexture: VideoTexture | null;
```
samjgorman, line 26: Can videoTexture be undefined? If so, we can refactor to 

````typescript
videoTexture?: VideoTexture

````

File: src/scene/players/PeerPlayerEntity.tsx [hunk 6]
```diff
@@ -92,43 +90,39 @@ const PeerPlayerEntity = ({
...
-      const newPosition = new Vector3(...Object.values(movementData.position));
-      playerVisualMesh.current.position.copy(newPosition);
-
-      model.current.position.addVectors(newPosition, offset);
-
-      playerCollisionMesh.current.position.copy(newPosition);
-
-      // Refit the collider boundsTree
-      playerCollisionMesh.current.geometry.boundsTree?.refit();
+    offset.fromArray([0, Math.sin(totalElapsedtime.current) / 30, 0]);
```
samjgorman, line 101: Why `/30`?

src/scene/players/PeerPlayerEntity.tsx
```diff
@@ -162,14 +156,14 @@ const PeerPlayerEntity = ({
...
-          <meshStandardMaterial attach="material-5" map={videoRef.current} />
+          <meshStandardMaterial attach="material-5" map={videoTexture} />
         </>
       </mesh>
       <group ref={model} />
       <mesh ref={playerCollisionMesh}>
         <meshBasicMaterial
           transparent={true}
-          opacity={0.0}
+          opacity={1.0}
```
samjgorman, line 166: Why did we change from 0 to 1? Is this fixing a previously uncaught bug with rendering wireframes when wireframe is set to true?

src/scene/players/PeerPlayerMediaProvider.tsx
```diff
@@ -35,7 +41,8 @@ const PeerPlayerMediaProvider = ({
 
   // Create positional audio object
   const audioRef = useRef<PositionalAudio>(new PositionalAudio(listener));
-  const videoRef = useRef<VideoTexture>();
+  const [videoTexture, setVideoTexture] = useState<VideoTexture | null>(null);
+  // const videoRef = useRef<VideoTexture>();
```
samjgorman, line 45: 👀 Let's remove this no longer used videoRef code in this file

src/scene/players/PlayerDataProvider.tsx
```diff
@@ -19,8 +17,10 @@ import UserPlayerEntity from './UserPlayerEntity';
 const PlayerDataProvider = () => {
   // Define state variables and refs for managing players
   const listenerRef = useRef(new AudioListener()); // Manages user audio listener
+  const playerDataMapRef = useRef(new Map<string, PlayerMovementData>());
+  const playerCollisionMapRef = useRef(new Map<string, Mesh>());
 
-  const setPeerMovementData = useSetAtom(playerMovementDataMapAtom);
+  // const peerMovementData = useAtomValue(playerMovementDataMapAtom);
```
samjgorman, line 23: 👀

src/scene/players/UserPlayerEntity.tsx
```diff
@@ -204,27 +210,30 @@ const UserPlayerEntity = ({ listener }: UserPlayerEntityProps) => {
...
-
       // adjust player position based on collisions with world
-      const collisionDisplacementVector = calculatePlayerCollisions({
-        playerCurrentPosition: playerColliderRef.current.position,
-        playerRadius: userData.radius,
-        playerRay: userData.ray,
-        playerWorldMatrix: playerColliderRef.current.matrixWorld,
-        boundsTree: meshCollider.geometry.boundsTree,
-      });
+      collisionDisplacementVector.fromArray([0, 0, 0]);
```
samjgorman, line 214: Given that we have this in 2 places may be helpful to refactor to constant

src/scene/players/UserPlayerEntity.tsx
```diff
@@ -204,27 +210,30 @@ const UserPlayerEntity = ({ listener }: UserPlayerEntityProps) => {
...
+          boundsTree: meshCollider.geometry.boundsTree,
+        },
+        collisionDisplacementVector,
+      );
 
       // adjust player position based on collisions with other peers
-      for (let [, peerMesh] of peerCollider) {
-        collisionDisplacementVector.add(
-          calculatePlayerCollisions({
+      for (let [, peerMesh] of playerCollisionMapRef.current) {
```
samjgorman, line 227: Style: Would prefer forEach / functional style over iterative loops...

````typescript
playerCollisionMapRef.current.forEach(([, peerMesh]) => {
  calculatePlayerCollisions(
    {
      playerCurrentPosition: playerColliderRef.current.position,
      playerRadius: playerColliderRef.current.userData.radius,
      playerRay: playerColliderRef.current.userData.ray,
      playerWorldMatrix: playerColliderRef.current.matrixWorld,
      boundsTree: peerMesh.geometry.boundsTree,
    },
    collisionDisplacementVector,
  );
});
````

File: src/common/utils/playerDynamics.ts [hunk 12]
```diff
@@ -57,32 +66,47 @@ type PlayerCollisionParams = {
...
 /**
  * Determine how to adjust the player when colliding with an object
  * @param params PlayerCollisionParams detailing required components to calculate player collision
  * @returns A unit vector pointing in the direction of adjustment if a collision is registered
  */
 export const calculatePlayerCollisions = (
   params: PlayerCollisionParams,
-): Vector3 => {
+  target: Vector3,
+): void => {
```
samjgorman, line 86: Shouldn't be necessary to explicitly type as `void`

```suggestion
) => {
```

src/common/utils/playerDynamics.ts
```diff
@@ -152,17 +169,18 @@ export const getUpdatedGravityVector = (
   displacementVector: Vector3,
   currentTickGravity: Vector3,
   isPlayerOnGround: boolean,
-): Vector3 => {
+  target: Vector3,
+): void => {
```
samjgorman, line 173:
```suggestion
) => {
```

src/scene/players/PeerPlayerEntity.tsx
```diff
@@ -73,9 +70,10 @@ const PeerPlayerEntity = ({
   // Update the player mesh atom to make sure we can globally access
   // this object
   useEffect(() => {
-    setPlayerMap(
-      (prev: any) => new Map(prev.set(peerId, playerCollisionMesh.current)),
-    );
+    playerCollisionMapRef.current.set(peerId, playerCollisionMesh.current);
+    return () => {
+      playerCollisionMapRef.current.delete(peerId);
```
samjgorman, line 75: Nice addition of a cleanup function here! :)

samjgorman (commented):
The memory leak is vanquished! 🐉 ⚔️ 

Left a round of style suggestions + questions

src/common/utils/playerDynamics.ts
samjgorman: Discussed offline that if we prefer, we can name this `GlobalCollisions` or similarly drop the props

src/common/utils/playerDynamics.ts
kanyesthaker: Whenever we apply collisions, we need to recompute the player position bounding box, since it the axis-aligned bounding box changes depending on the player position and orientation. To do this we create a box of size 0 (represented by the makeEmpty call) and morph it to the correct size.

src/common/utils/playerDynamics.ts
kanyesthaker: I think constants like 0 and 1, and related equivalents (i.e. [0, 0, 0] is equivalent to 0 in this case) are generally exempt from constant rules. See an example here, though not a hard and fast rule https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/MagicNumberCheck.html

kanyesthaker: will leave the fix for having a better default player position to a future PR

src/scene/players/PeerPlayerEntity.tsx
kanyesthaker: I don't think undefined is allowed. null is accepted by the downstream consumer, while undefined is not. Upstream of this file we only allow either VideoTexture or null as the type of this object.

src/scene/players/PeerPlayerEntity.tsx
kanyesthaker: We decided this when we were tuning constants. I can make this a constant since it is somewhat arbitrary.

src/scene/players/PeerPlayerEntity.tsx
kanyesthaker: mistake, fixed

src/scene/players/PeerPlayerMediaProvider.tsx
kanyesthaker: good eye

src/scene/players/UserPlayerEntity.tsx
kanyesthaker: see above. Don't think 0 is worth a constant, makes it harder to read + understand, which should be the ultimate goal

src/scene/players/UserPlayerEntity.tsx
kanyesthaker: thanks!

src/scene/players/UserPlayerEntity.tsx
```diff
@@ -224,18 +224,22 @@ const UserPlayerEntity = ({
...
-            playerRadius: playerColliderRef.current.userData.radius,
-            playerRay: playerColliderRef.current.userData.ray,
-            playerWorldMatrix: playerColliderRef.current.matrixWorld,
-            boundsTree: peerMesh.geometry.boundsTree,
-          },
-          collisionDisplacementVector,
-        );
-      }
+
+      // TODO (Kanyes): The memory leak refactor breaks peer-peer collisions. Commenting this out for now
```
samjgorman, line 228: Thanks for the TODO!

yarn.lock
samjgorman, line 1: These updates to yarn.lock seem unrelated to your change. Can we avoid checking this in? If you need to run yarn install, you can run `yarn install --frozen-lockfil`e to avoid updating the `yarn.lock`

samjgorman (approved):
Please remove the changes to yarn.lock and we're good to go!  Nice work

PR #237, opened 2023-07-06
fix: player model should no longer ever be stuck in the floor

### LINKED ISSUE #234: When peer is away from their tab, and another peer joins, the first peer will appear like they are stuck at the spawn position.

kanyesthaker:
### Description of Changes
Peers will get de-synced with other peers if no coordinates are being streamed, for example when the player has left their tab. This causes weird and unintuitive behavior, as the player's model would get set in an unnatural default animation loop. This fixes the issue by making sure that peers are always streaming coordinates. There should be no additional performance cost, as there is no extra computation being done per-frame.

### Link to Issue
closes #234 

### Testing Procedure
Tested by leaving one peer in a window, moving away from that window, and having another peer join. The original peer is in the appropriate position.


samjgorman (approved):
Looks good -- discussed async whether this could introduce more memory issues in the short term as the peer receives more stateful updates from their onMessage. 

We'll monitor for issues but unlikely as of today this would cause significant issue

PR #238, opened 2023-07-07
infra: added sentry sourcemap uploads

shalins:
### Description of Changes
* added sentry sourcemaps as part of build & deploy process


spacebar
```diff
@@ -1,12 +1,57 @@
...
+	fi
+
+	echo "Updated from v$VERSION_NUMBER to $INCREMENTED_VERSION"
+}
+
+function release_to_sentry() {
+	local service="$1"
+
+	ORG=$SENTRY_ORG
+  if [ "$service" == "client" ]; then
```
samjgorman, line 40: What if the service is `all`?  Seems like we'd then just set env vars for the server

.env.local.sample
```diff
@@ -1,5 +1,11 @@
+# These keys must be filled in order to upload the sourcemaps to Sentry properly
```
samjgorman, line 1: Why don't we commit these keys to the repo? Is this because we have multiple sentry orgs currently?

samjgorman (commented):
Nice work here! Left some questions for my own understanding + for clarity -- excited for this!

spacebar
```diff
@@ -28,6 +74,7 @@ Options:
     lint            Run linters on all files. Run with --fix to fix.
     logs            Stream most recent logs. Run with --client for client logs, --server for server logs.
     deploy          Deploy the client with --service client, the server with --service server, or both with --service all.
+    cancel          Cancel the client with --service client, the server with --service server, or both with --service all.
```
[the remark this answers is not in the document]
shalins: SIGINT just stops printing the build log to STDOUT, it doesn't kill the process. Cancel kills the process so you can actually cancel builds in the middle

spacebar
shalins: if you look at the flag parser logic, `all` just calls the spacebar script twice, once for client and once for server, so it should always go through these correctly

.env.local.sample
shalins: because it's generally bad practice to commit our api secret keys to the repo

server/app.ts
```diff
@@ -1,10 +1,17 @@
...
 import log from 'loglevel';
 import ws from 'ws';
 
 import { SignalingServer } from './server';
 import { DefaultPingInterval, PongOp } from './utils/consts';
 
+Sentry.init({
+  dsn: 'https://3141d7a37aa64506b6a4d9d8498e1c0e@o4505282763161600.ingest.sentry.io/4505486147387392',
+  release: process.env.SERVER_RELEASE_VERSION,
+  tracesSampleRate: 1.0,
```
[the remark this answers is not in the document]
shalins: sample rate just means rate at which we sample sentry errors, which is another way of saying what % of errors are captured? 1.0 means 100% of the logs are captured. most logging or error tracing tools do have some overhead associated that affect performance, but I think it it negligible compared to the other performance things we could optimize

server/package.json
```diff
@@ -4,24 +4,27 @@
...
-    "@types/webrtc": "^0.0.33",
+    "@types/webrtc": "^0.0.35",
     "express": "^4.17.3",
     "loglevel": "^1.8.0",
     "nodemon": "^2.0.15",
     "react-scripts": "^5.0.1",
     "simple-peer": "^9.11.1",
     "ts-node": "^10.9.1",
-    "typescript": "^4.6.3",
+    "typescript": "^5.1.6",
```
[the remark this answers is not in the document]
shalins: Is there a strong reason not to upgrade? The types were updated because there were some issues running the `tsc` command needed for building source maps with the version of the types we had.

.env.local.sample
samjgorman: Forgot this is public repo within a private team, this is fine then!

server/package.json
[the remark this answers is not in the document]
samjgorman: This should be fine -- we'll just keep an eye out for breaking changes

server/app.ts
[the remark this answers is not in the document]
samjgorman: Makes sense, especially considering we do this on client as well.  If we're seeing high volume of errors materially affect perf can revisit this var

spacebar
```diff
@@ -125,14 +173,38 @@ case $main_param in
...
                 ./spacebar deploy --service server; ./spacebar deploy --service client
             else
                 echo "Invalid service $service_name, must be one of (client|server)."
             fi
         fi
         ;;
-esac
+		cancel)
+        if [[ -z "$service_name" ]]; then
+            echo "Must specify a service with --service (client|server)."
```
samjgorman, line 196:
```suggestion
            echo "Must specify a service with --service (client|server|all)."
```

spacebar
```diff
@@ -125,14 +173,38 @@ case $main_param in
...
+                cd server
+								increment_version "spacebar-5"
+								. $ENV_LOC # Source to update the version in the current shell.
+								release_to_sentry "$service_name"
+								heroku builds:create -a spacebar-5 
+								cd ..
             elif [[ "$service_name" = "all" ]]; then
                 ./spacebar deploy --service server; ./spacebar deploy --service client
             else
                 echo "Invalid service $service_name, must be one of (client|server)."
```
samjgorman, line 190:
```suggestion
                echo "Invalid service $service_name, must be one of (client|server|all)."
```

samjgorman (approved):
this is it! 🤌

PR #243, opened 2023-07-25
fix: Adjust AFK behavior

### LINKED ISSUE #242: Player interactions are weird and unintuitive when one or both players are away from the active tab

kanyesthaker:
### Description of Changes
SpaceBar behaves very oddly when a player is away from their visible tab. Browser tabs will pause rendering for tabs not in focus, which messes with the render loop. Furthermore, this PR makes it so that players will spawn in a reasonable location with reasonable animations even if there is no input coming from the peer due to them being AFK.

### Link to Issue
closes #242 

### Testing Procedure
Will need to test live


src/scene/players/UserPlayerEntity.tsx
```diff
@@ -177,15 +175,29 @@ UserPlayerEntityProps) => {
...
 
   // Update movement vectors based on camera location
-  let tickDelta = 0;
-  const cameraAngle = new Quaternion();
-  const playerDirectionUnitVector = new Vector3();
-  const playerControlsVelocity = new Vector3();
-  const collisionDisplacementVector = new Vector3();
   useFrame((_, delta) => {
+    // If the frame delta is > 500ms, do not execute the logic below (useful when coming out of pause)
+    if (delta > 500) {
```
shalins, line 198: note: this will also trigger if the frame rate drops significantly. just double check that this is still want we want in that case? if not, increase the delta

src/scene/players/PeerPlayerEntity.tsx
```diff
@@ -119,24 +128,19 @@ const PeerPlayerEntity = ({
 
     if (position) {
       nextFramePosition.fromArray(Object.values(position));
-      playerVisualMesh.current.position.copy(nextFramePosition);
-      playerCollisionMesh.current.position.copy(nextFramePosition);
-      model.current.position.copy(nextFramePosition);
-      // Refit the collider boundsTree
-      playerCollisionMesh.current.geometry.boundsTree?.refit();
```
shalins, line 126: why was this code removed? it's not super clear to me why we no longer need this

src/scene/players/UserPlayerEntity.tsx
shalins: it would be useful to clarify why 500 was chosen and maybe define it as a constant

src/scene/players/PeerPlayerEntity.tsx
```diff
@@ -103,7 +112,7 @@ const PeerPlayerEntity = ({
 
   // On each frame, update the position of both the visual and collider mesh
   const offset = new Vector3();
-  const nextFramePosition = new Vector3();
+  const nextFramePosition = new Vector3(0, PlayerConstants.playerHeight, 0);
```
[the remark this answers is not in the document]
kanyesthaker: good point

src/scene/players/UserPlayerEntity.tsx
```diff
@@ -177,15 +175,29 @@ UserPlayerEntityProps) => {
     camera.add(listener.current);
   }
 
+  // Halt all movement if the tab is not in focus
+  window.addEventListener('blur', () => {
+    log.debug('Movement halted since tab is out of focus');
+    isMovingForward = false;
+    isMovingLeft = false;
+    isMovingBackward = false;
+    isMovingRight = false;
```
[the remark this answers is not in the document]
kanyesthaker: what do you mean by mutual exclusivity? A player can be moving in multiple directions simultaneously

src/scene/players/UserPlayerEntity.tsx
kanyesthaker: 500 is arbitrary. Frame rate dropping under 2FPS would be pretty devastating regardless, but I can bump this up to 2s so it won't be coupled w/ frame rate drops. Will make constant

src/scene/players/PeerPlayerEntity.tsx
kanyesthaker: just moved outside the if statement so it works reliably even when a player is on another tab (otherwise it is reset back to zero, this more or less "caches" the player's last seen position). The boundstree refit is only important if the base geometry morphs, which it does not.

shalins: Yeah I get that, I should've specified I meant the `refit` function.

src/scene/players/UserPlayerEntity.tsx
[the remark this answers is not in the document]
shalins: I guess we can keep for now, but 4 bools makes it also easy for state to get incorrect -> isMovingForward == true and isMovingBackward == true. Enums are variants are by definition mutually exclusive, so things that should never be set together aren't. How I'd write this would probably be with 2 enums, each with the variants that should be mutually exclusive to keep the code definitions precise. However I'll leave it up to you.

src/scene/players/PeerPlayerEntity.tsx
```diff
@@ -55,6 +55,10 @@ const PeerPlayerEntity = ({
   const playerVisualMesh = useRef<Mesh>(new Mesh());
   const playerCollisionMesh = useRef<Mesh>(new Mesh());
 
+  playerVisualMesh.current.position.setComponent(
+    1,
+    PlayerConstants.playerHeight,
+  );
```
shalins, line 61: nit: imo doing a setComponent and just updating the y-component feels a bit weird to me. Can we maybe switch to just doing `playerCollisionMesh.current.position.copy(PlayerConstants.playerSpawnPosition)`?

src/common/constants.ts
```diff
@@ -117,14 +117,18 @@ export abstract class SkyboxImages {
...
 
-  static readonly playerSpawnPosition: [number, number, number] = [0, 3, 0];
-
   static readonly movementSpeed = 5;
 
   static readonly jumpSpeed = 6;
 
   static readonly playerHeight = 2;
 
+  static readonly playerSpawnPosition: [number, number, number] = [
```
shalins, line 126:
```suggestion
  static readonly height = 2;

  static readonly spawnPosition: [number, number, number] = [
```

more clean imo to do PlayerConstants.spawnPosition -- no need for player twice. applies everywhere

shalins, line 126: why not just save this as a Vector3? if the reasoning is bc it requires an import, i actually think we should import it since i think "casting" it to a Vec3 everywhere it's used is uglier and also sorta defeats the point of a ready-to-use constant

src/common/constants.ts
```diff
@@ -140,6 +144,8 @@ export abstract class PlayerConstants {
   );
 
   static readonly playerIdleAnimationMagnitude = 30;
+
+  static readonly playerPhysicsTimeout = 2000;
```
shalins, line 148:
```suggestion
  static readonly playerPhysicsTimeout = 2000; // milliseconds
```

src/scene/players/UserPlayerEntity.tsx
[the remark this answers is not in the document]
kanyesthaker: I think it's ok for isMovingForward to be true and isMovingBackward to both be true, and have the net movement be zero. I think this is a pretty common pattern that you'll see in a lot of online games as well, or games that use the arrow keys (I actually installed minecraft just to check this, and that's how minecraft works. Shell shockers works the same way. Console games don't do this because a joystick is strictly mutually exclusive, with far more than 4 possible states).

src/common/constants.ts
kanyesthaker: Sure, we can. The purpose behind avoiding the import isn't for ugliness/cleanliness, it's more for safety (if Threejs changes how the initialize Vector3, the _constants_ file shouldn't be the one impacted). I don't think it makes a difference in this example, or because our codebase is so small. But I think it's a good general principle to keep in mind -- if we change from Threejs to some other framework, a triplet to describe space coordinates will still exist, but the Vector3 class may not. Somewhat similar to this http://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html

kanyesthaker: Good point. I can change this universally

src/scene/players/PeerPlayerEntity.tsx
kanyesthaker: sure, this is more consistent

src/scene/players/UserPlayerEntity.tsx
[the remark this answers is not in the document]
kanyesthaker: I also wonder what the interaction would be otherwise. Forward + Left = Forward-Left, Forward + Right = Forward-Right, Backward + Left = Backward-Left, Backward + Right = Backward-Right -- But would Forward+Backward keep going Forward? Or would it switch to backward? Either way, it breaks the pattern of movement directions adding up, and it could feel like certain movement keys are being disabled at random. I think the movement cancelling out is a more intuitive behavior. If we're curious about this, I'm happy to prototype on a test branch and learn from doing

src/common/constants.ts
kanyesthaker: Actually, I take that back -- I don't think it's safe to do this here. Right now, it could be OK because these values are read-only, and we copy them into a new Vector3 object every time they are used (we have to, because they are primitives). However, say we change it to a Vector3 object.

Suppose (this is actually pretty likely) we change the spawn position based on how many peers are in the room so that peers don't spawn on top of each other. If it's a Vector3, we might be inclined to use the spawnPosition object directly, or modify its values (making it static readonly does not prevent this). But doing so would impact all logic that depends on spawnPosition.

The only way around that would be to always use `const thisComponentSpawnPosition = new Vector3().copy(PlayerConstants.spawnPosition)` which is more difficult to reason about imo than instantiating the Vector3 with the primitive array directly.

shalins (approved):

PR #250, opened 2023-07-31
[Bug] Allow dynamic height resizing of canvas

samjgorman:
### Description of Changes

- We rely on` window.innerHeight `in order to set the height of the canvas. However, this is set once on component mount and then is unresponsive to future changes in height. This leads to the bug we've observed of the canvas not resizing correctly after resizing the browser window.
- This introduces a new common hook, `useWindowSizeObserver`, to observe changes in window size and return state that can be used by components
- Now, we use this hook to get the always up-to-date window height and can resize spacebar with abandon! 

### Link to Issue
https://github.com/IrisHub/SpaceBar/issues/251

### Testing Procedure
Manual QA by reloading spacebar with small height screen and then resizing the screen to be different heights and widths


src/scene/Scene.tsx
```diff
@@ -27,9 +27,11 @@ const Scene = () => {
   log.debug(
     `Pointer lock movement speed set to ${PlayerConstants.pointerSpeed}`,
   );
+
+  const { height } = useWindowSizeObserver();
+
   return (
-    <div style={{ height: window.innerHeight }}>
-      <Display />
```
samjgorman, line 32: We were rendering <Display/> twice, one in `<App/>` and here 😬

src/scene/Scene.tsx
```diff
@@ -27,9 +27,11 @@ const Scene = () => {
   log.debug(
     `Pointer lock movement speed set to ${PlayerConstants.pointerSpeed}`,
   );
+
+  const { height } = useWindowSizeObserver();
```
kanyesthaker, line 31: Should window size monitoring be at the highest level (ie in <App/>)? otherwise we might see strange behavior in the future, for example the pause menu, if it's not a child of <Scene/>

src/scene/Scene.tsx
kanyesthaker: good catch

src/scene/Scene.tsx
```diff
@@ -27,9 +27,11 @@ const Scene = () => {
...
+
   return (
-    <div style={{ height: window.innerHeight }}>
-      <Display />
+    <div style={{ height: height }}>
```
kanyesthaker, line 34: Do we not need to also set the width?

src/scene/Scene.tsx
samjgorman: The logic in `useWindowSizeObserver` is independent of any component in the tree, so I'm having a harder time seeing a case where this could cause the kind of case you describe.  I'd prefer to keep in its existing position and then when future components need access to height or width then they can either use this hook or get access via props

src/scene/Scene.tsx
samjgorman: Good question! You'd think we need to, but we don't in order to achieve our desired functionality. 

I'm actually not that sure why this is the case.  Possibly because the default width is the length of the screen while the default height is smaller than the height of the browser window, leading to us needing to override with this variable.  

We could add `width`, but would rather not if not required for functionality

src/scene/Scene.tsx
kanyesthaker: ah you're right, works for me

src/scene/Scene.tsx
kanyesthaker: makes sense, and tracks with what i've seen. Have never seen width be a problem. I'm fine with this explanation

kanyesthaker: Can we verify that hypothesis?

kanyesthaker (approved):

src/scene/Scene.tsx
samjgorman: Verified async & discussed

PR #257, opened 2023-08-06
Better dependency management

samjgorman:
### Description of Changes

- Move many dependencies that were incorrectly included as part of runtime `dependencies` back to `devDependencies` across Spacebar client + server
- Deleted unimported code that imported unused deps such as react/cannon
- Reduced Heroku slug size from `840mb` to `163.5mb` with the following optimizations: 
     -- Introduced a .`slugignore` to exclude node_modules and other unnecessary files
     -- Generating a fresh install of dependencies before building sourcemaps in the Spacebar deployment script 
     --  Excluding dirs such as `node_modules` in a .`slug-post-clean` file after using a custom buildpack that deletes assets used to generate a production build but not required as part of the slug
- Bonus -- Add Jotai debugging libraries for use in dev

### Link to Issue
Link the appropriate GitHub Issue for this PR. You can link an issue by adding `closes #issue-number` to the PR message body.

### Testing Procedure
yarn start & yarn build locally run as expected with no errors for client & server
./spacebar deploy -s client & ./spacebar deploy -s server run as expected & deploy successfully


.slugignore
```diff
@@ -0,0 +1,42 @@
+# This file specifies patterns that will be ignored
```
kanyesthaker, line 1: I'm unclear on what the difference between the .slugignore and .slug-post-clean. From a very quick google it looks like the .slugignore stops certain files/directories from even being pushed to heroku (before the buildpack runs), while .slug-post-clean removes files after the build is done. In that case, why would we need /.dvc and /data in the .slug-post-clean? shouldn't excluding them in the .slugignore be sufficient?

.slugignore
```diff
@@ -0,0 +1,42 @@
...
+yarn-debug.log*
+yarn-error.log*
+
+# Development and Debugging Files
+.env
+.env.local
+
+# Documentation and Source Files
+docs
+*.blend
```
kanyesthaker, line 21: instead of trying to catch all potential files, i think we should just have a /sandbox directory that is .gitignore'd and .slugignore'd that can contain all files that we don't want to ever touch prod. FYI this is what we do at work too

.slugignore
```diff
@@ -0,0 +1,42 @@
...
+*.md
+*.rst
+*.dvc
+
+
+# Sourcemaps
+.js.map
+
+# Testing and Debugging Files
+test
```
kanyesthaker, line 31: do we have any of these files?

package.json
```diff
@@ -77,16 +59,32 @@
...
-    "@typescript-eslint/parser": "^6.2.1",
+    "@types/three": "^0.155.0",
+    "@types/uuid": "^9.0.0",
+    "@types/ws": "^8.5.5",
+    "@typescript-eslint/eslint-plugin": "^5.13.0",
+    "@typescript-eslint/parser": "5.6.0",
+    "@typescript-eslint/typescript-estree": "6.2.1",
     "autoprefixer": "^10.4.13",
-    "eslint": "^8.46.0",
+    "eslint": "7.32.0",
```
kanyesthaker, line 85: why do we need to pin eslint at 7.32?

package.json
```diff
@@ -97,11 +95,14 @@
     "husky": "^8.0.3",
     "import-sort-cli": "^6.0.0",
     "import-sort-parser-typescript": "^6.0.0",
+    "jest": "^29.6.2",
```
kanyesthaker, line 98: nice, jest is heavy. not sure why this was in prod deps in the first place

spacebar
```diff
@@ -54,6 +54,9 @@ function release_to_sentry() {
 		TOKEN="$SENTRY_AUTH_TOKEN_SERVER"
 	fi
 
+    rm -rf node_modules
```
kanyesthaker, line 57: :thinking: this will remove my node_modules entirely, right? i.e. it's not doing this ephemerally, it is actually wiping/rebuilding my node modules? i.e. if i try to exit this flow midway through, i will simply no longer have my node_modules anymore?

spacebar
```diff
@@ -182,6 +185,7 @@ case $main_param in
 								increment_version "spacebar"
 								. $ENV_LOC # Source to update the version in the current shell.
 								release_to_sentry "$service_name"
+                heroku builds:cache:purge -a spacebar --confirm spacebar
```
kanyesthaker, line 188: what is builds:cache:purge? and what is this --confirm flag doing? What happens if a user exits this script after this line is run but before the next one?

kanyesthaker (commented):
thanks for doing this, some questions

package.json
samjgorman: @shalins do you have a POV on why we did this? This was rolled over from your initial PR.  We can also bring ESLint back to `8.46.0 ` and test if we still hit the blocking issue as this could have been noise. Otherwise, my best explanation would be that from using the no longer maintained create react app framework, upgrading packages can break build and runtime in non-obvious ways

shalins:
Closing this because #262 ended up using this branch as a starting point to address the unaddressed PR comments as well as implement some of the functionality that was missing in order to bring the changes proposed here to deployment.

shalins:
Closing this because #262 ended up using this branch as a starting point to address the unaddressed PR comments as well as implement some of the functionality that was missing in order to bring the changes proposed here to deployment.

PR #261, opened 2023-08-29
feat: enable basic reflections and player model pre-caching

### LINKED ISSUE #260: Support reflectivity in reflective meshes within the scene

kanyesthaker:
### Description of Changes
- Enables basic reflections by applying an environment map to all meshes in the scene.
- Upgrades the player model to a higher-resolution version that maintains reflectivity.
- Adds pre-caching for player model to prevent intermediate loading.

Warnings:
- This may cause some slight performance issues in the world. Best to play around with this, especially on different hardware, before merging.
- The reflections do not capture objects in the world, only the skybox. Will try to fix this in this PR but might bet bumped.

### Link to Issue
closes #260 

### Testing Procedure
image
image
image


src/scene/players/PeerPlayerEntity.tsx
```diff
@@ -84,16 +85,19 @@ const PeerPlayerEntity = ({
...
-      model.current.add(
-        getProcessedGLTFMeshes(res, {
-          position: new Vector3(0, 0, 0),
-          scale: 1.5,
-          quaternion: new Quaternion(0, 1, 0, 0),
-        }),
-      );
+    const modelEnv = getProcessedGLTFMeshes(gltf, {
+      position: new Vector3(0, 0, 0),
+      scale: 2.5,
```
kanyesthaker, line 95: note: bc I switched out the player model, the scale is different. We can fix this in Blender so that the scale is 1:1.

src/common/constants.ts
```diff
@@ -2,6 +2,14 @@ import log from 'loglevel';
 
 import { Bucket, resolveStoragePath, resolveStoragePaths } from './utils/io';
 
+export abstract class Graphics {
+  static readonly reflectionTargetBufferSize = 1024;
```
shalins, line 6:
```suggestion
  static readonly reflectionTargetBufferSize = 1024; // pixels
```

src/common/constants.ts
```diff
@@ -2,6 +2,14 @@ import log from 'loglevel';
...
+
+  static readonly reflectionNear = 1;
+
+  static readonly reflectionFar = 2048;
```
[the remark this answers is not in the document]
shalins: also any reason why the zNear is 1 meter? What happens if we make it 0.1?

src/common/utils/models.ts
```diff
@@ -1,25 +1,36 @@
...
   Quaternion,
+  Scene,
   Vector3,
+  WebGLCubeRenderTarget,
+  WebGLRenderer,
 } from 'three';
 import { MeshBVH, StaticGeometryGenerator } from 'three-mesh-bvh';
 import { GLTF } from 'three-stdlib';
 
+import { Graphics } from '../constants';
```
shalins, line 16: It would be better if this wasn't namespaced as `Graphics`, maybe something clearer like `Environment` or even `CubeCamera`

src/common/utils/models.ts
```diff
@@ -1,25 +1,36 @@
...
 /**
- * Reads the individual meshes within the import glTF, applies the appropriate parameters,
- * and returns the extracted meshes. These steps are essential to get mesh skinning to work properly
+ * Reads the individual meshes within the import glTF, applies the appropriate parameters, including
+ * position, scale, and the environment map for reflections, and returns the extracted meshes.
+ * These steps are essental to get mesh skinning to work properly.
  * @param gltf the glTF object loaded from a file
  * @param params GLTFMeshParams object defining the object position and scale
+ * @param gl The main renderer (for reflections)
+ * @param scene The main threejs scene (for refslections)
```
shalins, line 33:
```suggestion
 * @param gl Main renderer (for reflections)
 * @param scene Main threejs scene (for reflections)
```

src/common/utils/models.ts
```diff
@@ -39,6 +54,18 @@ export const getProcessedGLTFMeshes = (
       obj.receiveShadow = true;
       // Cull the object when it is out of view
       obj.frustumCulled = false;
+
+      // Apply the environment map for reflective surfaces
+      // We need a new camera located at the center of each mesh
+      // so the reflections are from the perspective of that mesh
```
shalins, line 60:
```suggestion
      // TODO: Place the camera at the center of each mesh
      // so the reflections are from the perspective of that mesh
```

src/common/utils/models.ts
```diff
@@ -39,6 +54,18 @@ export const getProcessedGLTFMeshes = (
       obj.receiveShadow = true;
       // Cull the object when it is out of view
       obj.frustumCulled = false;
+
+      // Apply the environment map for reflective surfaces
```
shalins, line 58:
```suggestion
      // Applies the environment map for reflective surfaces
```

src/common/utils/models.ts
```diff
@@ -39,6 +54,18 @@ export const getProcessedGLTFMeshes = (
       obj.receiveShadow = true;
       // Cull the object when it is out of view
       obj.frustumCulled = false;
```
shalins, line 56: This is unrelated but the comment and code don't match here. a) Why is frustum culling turned off? b) The comment says it's culled when out of view but that's not true when it's turned off :)

src/scene/players/PeerPlayerEntity.tsx
```diff
@@ -84,16 +85,19 @@ const PeerPlayerEntity = ({
...
+      quaternion: new Quaternion(0, 1, 0, 0),
```
shalins, line 96: would comment that there are lot of magic numbers here, but you wrote it as a todo already so ¯\_(ツ)_/¯

src/scene/players/PeerPlayerEntity.tsx
```diff
@@ -191,3 +195,6 @@ const PeerPlayerEntity = ({
 };
 
 export default PeerPlayerEntity;
+
+// Pre-load the model, so that the model is cached and we don't have to wait on first load
+useGLTF.preload(PlayerConstants.model);
```
shalins, line 200: nice!

shalins (approved):
Left some style comments, but nice work!

src/common/constants.ts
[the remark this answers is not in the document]
kanyesthaker: i don't think there's any major loss in precision if you move the near plane to 0.1; I think I just used the Threejs tutorial defaults, but i think any value in (0, 1] should be fine.

src/common/utils/models.ts
kanyesthaker: good point, i wasnt too careful when I namespaced this. I think calling it ReflectionCamera would probably be best

src/common/utils/models.ts
kanyesthaker: really nice catch, I do this a lot, should be 3rd person over 2nd person

src/common/utils/models.ts
kanyesthaker: huh you're right, I think I switched it from `true` to `false` several months ago but forgot to edit the comment

src/scene/players/PeerPlayerEntity.tsx
kanyesthaker: yea these should be scrubbed soon. I don't think we need explicit constants for (0, 0, 0), i'm on the fence about the quat -- maybe worth doing, i'll think about it

src/common/utils/models.ts
kanyesthaker: this isn't actually a TODO, this is what it currently does (or it's what it should be doing, it's hard to tell if it works right now because there's no reference frame while the reflections only show the skybox). Let me edit the comment to make this clearer

src/common/utils/models.ts
kanyesthaker: This should be `true`. Not sure why I set it to false, can't find a good reason :thinking:

PR #262, opened 2023-09-04
deps: QOL improvements to project dependencies

shalins:
### Description of Changes
Take on from where Sam left off in #257:
* Clean up dependency versions. Removed everything that we were not using (using `depcheck` https://www.npmjs.com/package/depcheck). This reduced slug size to **114.6M** for client.
* Upgraded all of our packages to the latest version. Also upgraded yarn to yarn berry (https://yarnpkg.com/getting-started/install) which changes our linker from node_modules to plug n play which increases installation speed and cold start time. Made corresponding changes to `server` module too.
* Reduced Heroku RAM usage from 1200MB to 196MB, 6x improvement by serving static build instead of running react-scripts, which should avoid any swap usage and make spacebar feel much more "snappy" in production.
* Refactored the `spacebar` script to make it ~50 lines less with much better abstractions, and also added a `clean` command that accomplishes doing a destructive operation to clear local packages and the heroku cache. Split into separate PR at #264.


package.json
```diff
@@ -1,62 +1,42 @@
...
-    "build": "react-scripts --optimize_for_size --max_old_space_size=2048 build",
+    "start": "yarn serve -s build",
+    "localStart": "yarn react-scripts --optimize_for_size --max_old_space_size=460 --gc_interval=100 start",
+    "localBuild": "react-scripts --optimize_for_size --max_old_space_size=2048 --gc_interval=100 build",
     "format": "prettier --write \"**/*.{ts,tsx,json,md}\"",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "heroku-postbuild": "yarn localBuild"
```
shalins, line 39: This was changed so that we can run different things locally vs when deploying. For example, we want to use react scripts to run when running locally, but we want to serve statically when deployed. LMK opinions on what this command should be called if we don't like `yarn localStart` @samjgorman @kanyesthaker

package.json
```diff
@@ -1,62 +1,42 @@
 {
-  "name": "3dspatialaudio",
+  "name": "spacebar",
+  "license": "MIT",
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@fortawesome/fontawesome-svg-core": "^6.1.1",
-    "@fortawesome/free-solid-svg-icons": "^6.1.1",
+    "@fortawesome/fontawesome-svg-core": "^6.4.2",
```
shalins, line 7: all packages upgraded, most packages removed

.yarnrc.yml
```diff
@@ -0,0 +1,3 @@
+nodeLinker: pnp
```
shalins, line 1: with yarn berry, the convention is to commit the cache (ref: https://stackoverflow.com/questions/62984576/committing-yarn-directory-to-git-when-using-yarn-berry)

server/tsconfig.json
```diff
@@ -5,6 +5,7 @@
 		"target": "es6",
     "rootDir": "./",
 		"outDir": "./out",
+    "skipLibCheck": true, // TODO(Sam): Set back to false for stricter type safety https://github.com/IrisHub/SpaceBar/issues/259
```
shalins, line 8: @samjgorman why was this set to true?

tsconfig.json
```diff
@@ -7,7 +7,7 @@
 		"sourceRoot": "/",
 		"outDir": "./out",
     "allowJs": true,
-    "skipLibCheck": true,
+    "skipLibCheck": false,
```
shalins, line 10: @samjgorman same thing here

.slug-post-clean
```diff
@@ -0,0 +1,11 @@
+# This file specifies patterns that will be removevd
```
shalins, line 1: @samjgorman i want to get rid of this file since it doesn't actually do anything, but removing this makes the heroku builds:create fail. Did you add a plugin that makes this file mandatory?

kanyesthaker: there's a buildpack that requires this file (i think it's called react-post-clean or similar), you can remove it from the heroku console

samjgorman: Yup!

.pnp.loader.mjs
samjgorman, line 1: I'm concerned that using Yarn plug & play will introduce more complexity than our project needs right now. 

Why not use Yarn 3 with node_modules? See (https://v3.yarnpkg.com/getting-started/migration#why-should-you-migrate)

Here's what gives me hesitation with yarn PNP

1. Requires special SDKs & plugins to interface with in VSCode & other IDEs  https://yarnpkg.com/getting-started/editor-sdks
2. Stricter approach to unlisted dependencies that errors out when dependencies rely on unlisted other dependencies. This is more robust in practice, but I question whether this may be worth the time tradeoff.  It is my understanding that many large-scale packages rely on unlisted dependencies (https://yarnpkg.com/migration/pnp#fix-dependencies-with-packageextensions), which requires additional work to resolve and occasionally leads to just unsupported packages.  React Native & Flow are large examples
4. The main problem that yarn PNP solves appears to be reducing install time via removing node modules, which for a project of our size I wouldn't anticipate to be a large problem for some time. Disk space is definitely a concern but does deploying a static build accomplish an adequate reduction? 

See discussion: 
https://spin.atomicobject.com/2022/05/02/yarn-pnp/
https://news.ycombinator.com/item?id=24869475

Next steps: 
1. What are your thoughts here? Do the benefits outweigh tradeoffs? 
2. For this PR, what's the lift to use yarn 3 & node modules and then spin off the yarn p&p PR as separate such that we could easily revert if we decide to go back?

samjgorman (commented):
Looks great but questioning whether we should upgrade to Yarn PNP now vs continue with node_modules with yarn 3

.slug-post-clean
shalins: I already tried searching for that buildpack and it wasn't showing up for me, which is why i left this comment. But I'll check again, thanks!

.yarnrc.yml
shalins: changed to node_modules

samjgorman (approved):
Looks great! Excited to get this in!

server/tsconfig.json
samjgorman: This most likely no longer applies and can be removed -- I'm okay with this getting picked up in a later change though

samjgorman: This was initially set to true to workaround the issue we had hit where building in `server` hit build issues from dependencies contained within` /src`. See the linked issue for a longer description!

tsconfig.json
samjgorman: This was a case where we actually didn't need skipLibCheck to be true, so setting back to false increases our type safety such that we ensure we perform type checking over type declaration files. The only time we'd want to set this to true from my knowledge is to workaround build issues such as what we experienced above

https://www.typescriptlang.org/tsconfig#skipLibCheck

server/tsconfig.json
shalins: Ok cool, thanks for linking the issue and for the response, I will leave this in for now to avoid bundling lots of untested changes in a single PR, but will do a future cleanup.

tsconfig.json
shalins: gotcha, good to know!

package.json
shalins: Since no comment, am keeping it the same for now.

.pnp.loader.mjs
shalins: For documentation purposes, we ended up going with node_modules instead of PNP. @samjgorman is right that there are not enough benefits to taking on the additional burden of using PNP, which not all packages currently support.

package.json
```diff
@@ -77,31 +57,29 @@
...
+    "eslint-plugin-react": "^7.33.2",
+    "eslint-plugin-simple-import-sort": "^10.0.0",
     "husky": "^8.0.3",
-    "import-sort-cli": "^6.0.0",
-    "import-sort-parser-typescript": "^6.0.0",
-    "lint-staged": "^13.1.0",
-    "postcss": "^8.4.19",
-    "prettier": "^3.0.0",
-    "pretty-quick": "^3.1.3",
-    "tailwindcss": "^3.2.4"
```
samjgorman, line 104: Looks like this is where we uninstalled Tailwind unintentionally

PR #264, opened 2023-09-07
infra: spacebar script refactor

shalins:
### Description of Changes
* Refactored the spacebar script to make it ~50 lines less with much better abstractions, and also added a clean command that accomplishes doing a destructive operation to clear local packages and the heroku cache.

### Testing Procedure
* Everything seems to work as expected.


spacebar
```diff
@@ -3,214 +3,189 @@
...
-            ;;
-        *) 
-            positional_args+=("$1")
-            shift
-            ;;
+        -f|--fix) fix_flag=1; shift ;;
+        -s|--service) service_name="${2:-}"; shift 2 ;;
+        -v|--verbose) set -x; shift ;;
+        -h|--help) usage ;;
+        *) main_param="$1"; shift ;;
```
kanyesthaker, line 161: this behaves a bit differently from the old code. The idea in the old example was that we could take in a list of positional args and then process them, for example linting a specific set of files by glob i.e. `./spacebar lint --fix *.ts,*.tsx`. I'm fine with this change though since the intent never materialized.

spacebar
```diff
@@ -3,214 +3,189 @@
...
-                yarn eslint --fix $files
-                yarn prettier --write $files
-            else
-                yarn eslint $files
-                yarn prettier --check $files
-            fi
-        fi
+        [[ -z "$files" ]] && echo "No files to lint." && exit 0
+        echo "Linting files: $files"
+        cmd_args=$([[ "$fix_flag" -eq 1 ]] && echo "--fix" || echo "")
```
kanyesthaker, line 171: why the `cmd_args` substitution for eslint but not for prettier?

spacebar
```diff
@@ -3,214 +3,189 @@
...
-                echo "Setting client-level logs to debug"
-                heroku logs --app spacebar --tail
-         elif [[ "$service_name" = "server" ]]; then
-                heroku logs --app spacebar-5 --tail
-         else
-                echo "Invalid service $service_name, must be one of (client|server)."
-         fi
-    fi
-    ;;
+        [[ -z "$service_name" ]] && die "Specify a service with --service (client|server|all)."
```
kanyesthaker, line 176: nit: save the error message as a variable i.e. `service_selection_missing_msg="Specify..."`

spacebar
```diff
@@ -3,214 +3,189 @@
...
-								release_to_sentry "$service_name"
-								heroku builds:create -a spacebar-5 
-								cd ..
-            elif [[ "$service_name" = "all" ]]; then
-                ./spacebar deploy --service server; ./spacebar deploy --service client
-            else
-                echo "Invalid service $service_name, must be one of (client|server|all)."
-            fi
-        fi
+        [[ -z "$service_name" ]] && die "Specify a service with --service (client|server|all)."
```
kanyesthaker, line 180: not sure if important, but these will all fail silently if you pass in `-s` without an argument

spacebar
```diff
@@ -3,214 +3,189 @@
...
-                ./spacebar cancel --service server; ./spacebar cancel --service client
-            else
-                echo "Invalid service $service_name, must be one of (client|server)."
-            fi
-        fi
+    clean)
+        handle_service_operation perform_cleanup
         ;;
-	esac
+    *) die "Invalid command $main_param. Must be one of [lint, deploy, cancel, logs, clean]." ;;
```
kanyesthaker, line 190: nit: since this is getting unwieldy, can do something like `Type ./spacebar -h for a list of available commands`?

kanyesthaker (commented):
quick glance, nice cleanup

spacebar
```diff
@@ -3,214 +3,189 @@
...
 ENV_LOC="$PWD/.env.local"
 
-source $ENV_LOC
+# Ensure the .env.local file exists
+if [[ ! -f "$ENV_LOC" ]]; then
+    echo "No .env.local file found."
+    exit 1
+fi
 
+# shellcheck disable=SC1090
```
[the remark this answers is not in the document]
shalins: sh is using bourne shell not bash, even though when you run ./spacebar it will be running with bash as the interpreter because of the shebang. so sh -n won't catch some of the bash specific things. also fwiw, shellcheck is amazing, we should really be adding a commit hook that makes running it mandatory. the previous script had many issues like unquoted variable expansions that shellcheck would have caught.

spacebar
```diff
@@ -3,214 +3,189 @@
...
+# Helper functions for the service operations
+heroku_logs() {
+    local app_name="$1"
+    echo "Streaming most recent logs for $app_name"
+    heroku logs --app "$app_name" --tail
+}
+
+deploy_app() {
+    local app_name="$1"
+    increment_version "$app_name"
```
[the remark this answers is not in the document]
shalins: good catch, fixed

spacebar
```diff
@@ -3,214 +3,189 @@
...
+    fi
+
+    sentry-cli --auth-token "$token" releases new "$version" --project "$project" --org "$SENTRY_ORG"
+    sentry-cli --auth-token "$token" releases files "$version" upload-sourcemaps --rewrite --org "$SENTRY_ORG" --project "$project" ./build/static/js
+}
+
+# Helper functions for the service operations
+heroku_logs() {
+    local app_name="$1"
+    echo "Streaming most recent logs for $app_name"
```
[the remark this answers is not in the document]
shalins: Agree! OOS for this PR though

spacebar
shalins: really good catch -- fixed

spacebar
```diff
@@ -3,214 +3,189 @@
...
 Options:
     -h --help       Show the help menu.
     -v --verbose    Print bash trace.
     lint            Run linters on all files. Run with --fix to fix.
     logs            Stream most recent logs. Run with --client for client logs, --server for server logs.
-    deploy          Deploy the client with --service client, the server with --service server, or both with --service all.
-    cancel          Cancel the client with --service client, the server with --service server, or both with --service all.
+    deploy          Deploy the client with --service client, server, or all.
+    cancel          Cancel the client with --service client, server, or all.
+    clean           Clean up node_modules, yarn cache, and Heroku build cache. WARNING: This is a destructive operation.
```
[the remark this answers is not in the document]
shalins: Simplified to just cleaning local yarn cache (https://classic.yarnpkg.com/lang/en/docs/cli/cache/#toc-yarn-cache-clean) and heroku builds cache (https://github.com/heroku/heroku-builds#purge-build-cache). Will extend as needed.

spacebar
shalins: Lets add back if needed.

kanyesthaker (approved):

spacebar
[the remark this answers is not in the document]
shalins: ok there was a reason. we're getting the number from heroku, so if i deploy first then it will pull the new version, increment it, and then update heroku again, leading to skipping a version each time. 

i.e. old workflow: -> read current heroku version -> 134 -> increment to 135 -> push to heroku
      with reordering: -> push to heroku -> heroku increments to 135 -> read heroku -> 135 -> increment to 136 -> push to heroku

PR #258, opened 2023-08-06
WIP pause screen

kanyesthaker:
### Description of Changes
Add a description of the changes you've made and would like to merge.
A quick sentence or bullet point for each point is sufficient -- this is to make it easy to understand changes to the codebase that your reviewer may not be familiar with.

### Link to Issue
Link the appropriate GitHub Issue for this PR. You can link an issue by adding `closes #issue-number` to the PR message body.

### Testing Procedure
Describe your testing procedure here. This can include unit tests, screen recordings of functionality, etc.

PR #265, opened 2023-09-12
Pause menu

samjgorman:
### Description of Changes

Stacked from https://github.com/IrisHub/SpaceBar/pull/258 @kanyesthaker, with the addition of...

- Fix for crash for peer upon changing audio 
- Sync state when spacebar keyboard shortcut is pressed
- More debug logging 

### Link to Issue
Link the appropriate GitHub Issue for this PR. You can link an issue by adding `closes #issue-number` to the PR message body.

### Testing Procedure


src/interface/VideoPlayer.tsx
```diff
@@ -44,139 +32,31 @@ const AVErrorMessages: Record<string, string> = {
...
-const VideoPlayer = ({ height, width }: VideoProps) => {
-  const [audioEnabled, setAudioEnabled] = useState(true);
-  const [videoEnabled, setVideoEnabled] = useState(true);
-  const [errMessageList, setErrMessageList] = useState<string[]>([]);
-  const [errMessage, setErrMessage] = useState<string | undefined>(undefined);
+const VideoPlayer = () => {
+  const isAudioEnabled = useAtomValue(isAudioEnabledAtom);
+  const isVideoEnabled = useAtomValue(isVideoEnabledAtom);
+  const isPaused = useAtomValue(isPausedAtom);
+  const [errMessage] = useState<string | undefined>(undefined);
```
kanyesthaker, line 39: Is errMessage ever set?

src/interface/components/MenuButton.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
```
kanyesthaker, line 1: Delete this file or implement it

src/interface/components/MenuContainer.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
```
kanyesthaker, line 1: Ditto

src/network/hooks/useWriteCommunicator.ts
```diff
@@ -0,0 +1,42 @@
+import { useAtomValue } from 'jotai';
```
kanyesthaker, line 1: I don't think I understand the purpose of the `useWriteCommunicator` and `useReadCommunicator` refactor. From reading through this code and the upstream in `PlayerDataProvider` it doesn't seem like there's a major functional difference. Is there a good reason to do this vs. keeping the callback definitions where they are? Might be missing something

src/network/hooks/useWriteCommunicator.ts
```diff
@@ -0,0 +1,42 @@
...
+        log.error(`Error sending stream: ${err}`);
+      }
+    }
+  }, [communicator?.isConnectedP2P, userMediaStream]);
+
+  return { sendData, streamMedia, initStreamMediaEffect };
+
+};
+
+export default useWriteCommunicatorData;
```
kanyesthaker, line 42: useWriteCommunicatorData --> useWriteCommunicator ?

src/scene/Scene.tsx
```diff
@@ -20,6 +23,10 @@ import PlayerDataProvider from './players/PlayerDataProvider';
  * a single element that can then be rendered in App.tsx
  */
 const Scene = () => {
+  const { restartUserMediaStream } = useUserMediaStream();
+
+  useEffect(() => restartUserMediaStream(true, true), []);
```
kanyesthaker, line 28: Let's add a comment --> "Need to request permissions on first load"

src/scene/players/PeerPlayerEntity.tsx
```diff
@@ -50,6 +52,13 @@ const PeerPlayerEntity = ({
   // Track total elapsed time for use in player animations
   const totalElapsedtime = useRef<number>(0);
 
+  // TODO: fix state update issue
```
kanyesthaker, line 55: What is "State update issue"? I don't remember if this is from my code or not, is this resolved?

src/scene/players/PeerPlayerMediaProvider.tsx
```diff
@@ -34,77 +40,58 @@ const PeerPlayerMediaProvider = ({
...
+      audioRef.current.setDirectionalCone(...AudioConstants.DirectionalCone); // Directional cone outside of which audio is muted
+
+      if (process.env.REACT_APP_SANDBOX_FLAG === 'true') {
+        const helper = new PositionalAudioHelper(audioRef.current, 3, 10, 5);
+        audioRef.current.add(helper);
       }
+      log.debug(`${Date.now()}: Set audio from peer stream`);
     }
-  }, [peerStream, isAudioAvailable, isVideoAvailable]);
+  }, [peerStream]);
```
kanyesthaker, line 94: Just want to double check the deps arrays here are correct, based on what we were playing around with with respect to updating on state attributes. I think this is fine for now since we currently replace the entire MediaStream every time we mute/toggle video. But in the very near future when we just want to modify tracks directly these will not fire.

src/scene/players/PlayerDataProvider.tsx
```diff
@@ -20,12 +21,16 @@ const PlayerDataProvider = () => {
   const playerDataMapRef = useRef(new Map<string, PlayerMovementData>());
   const playerCollisionMapRef = useRef(new Map<string, Mesh>());
 
-  // const peerMovementData = useAtomValue(playerMovementDataMapAtom);
   const setStreamMap = useSetAtom(playerStreamMapAtom);
   const otherPeersInRoom = useAtomValue(otherPeersInRoomAtom);
 
+  const { initStreamMediaEffect } = useWriteCommunicatorData();
```
kanyesthaker, line 27: See my comments above for useRead/WriteCommunicator

src/scene/players/UserPlayerEntity.tsx
```diff
@@ -342,49 +290,20 @@ UserPlayerEntityProps) => {
           w: camera.quaternion.w,
         },
       };
-      communicator?.send(playerMovementData);
+      sendData(playerMovementData);
```
kanyesthaker, line 293: Ditto -- is this really any more clear than what it was before?

kanyesthaker (commented):
Thanks for rounding this out. Some comments:
1. Let's add top-level docstrings for all the interface components
2. I'm unclear as to whether there's tangible benefit to the useCommunicator hook refactor. Could you explain this a bit more?

src/common/hooks/usePlayerControls.tsx
```diff
@@ -0,0 +1,160 @@
...
+  log.debug(`Player controls enabled set to ${!!isUserPlayerControlsEnabled}`);
+
+  // Handle pointer lock effects
+  useEffect(() => {
+    // Create the PointerLockControls instance
+    cameraControlsRef.current = new PointerLockControls(camera, gl.domElement);
+    cameraControlsRef.current.pointerSpeed = PlayerConstants.pointerSpeed;
+
+    if (isPaused) {
+      // Deactivate PointerLockControls
```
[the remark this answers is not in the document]
samjgorman: I actually think having the explicit connction from unlock -> deactivate pointerlock & lock -> activate pointerlock is helpful -- at least for me!  I'll vote for keeping these

src/common/hooks/usePlayerControls.tsx
```diff
@@ -0,0 +1,160 @@
...
+      window.removeEventListener('keydown', handleMovementKeyDown);
+      window.removeEventListener('blur', haltMovement);
+    };
+  }, [isPaused, isUserPlayerControlsEnabled]);
+
+  const handleKeyboardShortcuts = (e: KeyboardEvent) => {
+    switch (e.code) {
+      // Space -> Keyboard shortcut for toggling pause menu
+      case 'Space':
+        // The space key's default behavior triggers a click,
```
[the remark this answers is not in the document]
samjgorman: https://www.quirksmode.org/dom/events/click.html

> click
Fires when a mousedown and mouseup event occur on the same element.
Fires when the user activates the element that has the keyboard focus (usually by pressing Enter or the space bar).

src/interface/PauseScreen.tsx
```diff
@@ -0,0 +1,76 @@
...
+            <ambientLight intensity={1} />
+            <directionalLight position={[10, 10, 5]} intensity={1} />
+            <UserPlayerSelfModel />
+          </Canvas>
+        </div>
+        <div className="flex flex-row justify-center items-center gap-[24px] m-4">
+          {/* Button to toggle audio */}
+          <button
+            className={
+              'w-[120px] h-[60px] cursor-pointer hover:bg-opacity-50 p-2 bg-white bg-opacity-75 rounded-lg drop-shadow-md transition-all duration-200 ease-in-out'
```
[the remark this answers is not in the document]
samjgorman: Not worth doing in this PR, but agree that as components become more reusable, we want to centralize their styling so that we can enforce consistency 

We can also lint our Tailwind CSS via extensions to enforce patterns to make sure these style blobs are readable and predictable

src/common/utils/models.ts
```diff
@@ -67,7 +67,8 @@ export const getProcessedGLTFMeshes = (
       );
       obj.getWorldPosition(meshReflectionCamera.position);
       obj.material.envMap = cubeRenderTarget.texture;
-      meshReflectionCamera.update(params.gl, params.scene);
+      // TODO: Sam -- Clean up this type casting to be safer
```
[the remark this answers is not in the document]
samjgorman: Looks like we can avoid casting entirely and remove these so went ahead and did this --

any time we're type casting we're circumventing Typescript type-checking and opening ourselves up to more potential for bugs so in this case, type casting as Scene or WebGLRenderer doesn't actually guarantee that the values are of those types. Rather they could be undefined or subtly different and cause downstream bugs

src/interface/VideoPlayer.tsx
samjgorman: Hmm good catch! Looks like the refactor to useUserMediaStream missed adding error handling. I'll either add as part of this PR or track for a follow-on PR

samjgorman: OK, added error handling back to `useUserMediaStream`

src/interface/PauseScreen.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const [isVideoEnabled, setVideoEnabled] = useAtom(isVideoEnabledAtom);
+  const [isAudioEnabled, setAudioEnabled] = useAtom(isAudioEnabledAtom);
+  const { restartUserMediaStream } = useUserMediaStream();
+
+  return (
+    <PauseScreenBlocker>
+      <PauseMenuContainer>
+        <div className="pointer-events-none translate-x-6 w-[30vh] h-[30vh]">
+          <Canvas camera={{ fov: PlayerConstants.cameraFov }}>
+            <ambientLight intensity={1} />
```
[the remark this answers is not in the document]
samjgorman: Default for `ambientLight` is actually 1 so I'll remove, otherwise let's add constants

src/scene/players/PeerPlayerEntity.tsx
samjgorman: When developing locally I was hitting an uncaught exception on loading the state so marked as a TODO -- however looks like I'm no longer able to repro the issue. I'll monitor this & fix-forward if necessary, otherwise will remove the comment

src/network/hooks/useWriteCommunicator.ts
samjgorman: The goal: 
1. Burn down uses of `useEffect` in our codebase, which often fire in non-obvious ways & introduce error-prone, non-obvious code
2. Centralize write logic to communicator in a reusable hook that wraps base communicator methods with helpful logging & render optimization via useCallback with correct deps

So from that intention, I'm happy with the refactor as it achieved removing complex logic in a useEffect in userPlayerEntity, and made this component closer to becoming a presentational component that renders the user rather than also being responsible for maintaining the lifecycle of its data. It also made streaming media reusable for future components. 

I understand that for the `sendData` method, we don't see nearly the same degree level of value that we do from the `streamMedia`method. So from that respect it's a bit more of a shallow wrapper than I'd like. However, seems worth tradeoff for the streamMedia method. WDYT?

src/scene/players/PlayerDataProvider.tsx
samjgorman: See above

src/scene/players/UserPlayerEntity.tsx
samjgorman: See above -- we get the most value from `streamMedia` wrapper and agree that `sendData` isn't too impactful

src/common/hooks/usePlayerControls.tsx
```diff
@@ -0,0 +1,160 @@
...
+        // which can trigger clicks to toggling AV when we don't want to
+        // So, we prevent the event's default behavior
+        e.preventDefault();
+        setIsPaused(!isPaused);
+        break;
+      // M ->  Keyboard shortcut for muting
+      case 'KeyM':
```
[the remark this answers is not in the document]
samjgorman: @shalins and I QAd pause menu recently and both thought it would be helpful for us to be able to use keyboard shortcuts. Agree that v is less obvious and we'd want to iterate a bit. I'll remove the switch cases & add a TODO to add shortcuts

src/scene/players/PeerPlayerMediaProvider.tsx
samjgorman: Yeah including `isAudioAvailable` + `isVideoAvailable` was actually contributing to some buggy behavior for the receiving peer and don't need to be included in this implementation. Let's tackle updating this when we add follow-on work to solve the delay problem for peers when restarting AV

src/interface/components/HUDRowContainer.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+type HUDRowContainerProps = {
+  children?: ReactNode;
+};
+
+const HUDRowContainer = ({ children }: HUDRowContainerProps) => {
+  return (
+    <div
+      className={
+        'bg-white bg-opacity-50 rounded-lg m-2 p-2 flex flex-row justify-center items-center [&>*]:px-2'
```
samjgorman, line 11: @kanyesthaker I'd like to clean up this less-obvious part of CSS here` [&>*]:px-2` (was this GPT generated?) but am okay tackling in future work

kanyesthaker: got it from here! https://stackoverflow.com/questions/67119992/how-to-access-all-the-direct-children-of-a-div-in-tailwindcss

src/network/hooks/useWriteCommunicator.ts
kanyesthaker: I'm not sure if I'm following -- why do we need to burn down uses of useEffect? I understand that useEffect, like all other tools, can be dangerous if used incorrectly. Were we using it incorrectly? What was incorrect about the previous implementation?

Also, why do we need two different read and write hooks vs. just extending the existing useCommunicator hook? I find it hard to reason about what reading and writing means -- we aren't actually reading or writing any data (i.e. to disk) -- the separation in functionality between the two isn't clear to me

kanyesthaker (commented):
thanks for the fixes! Looks good, but have one more question around the communicator refactor. Let's discuss live and then i'll approve

kanyesthaker (approved):

PR #269, opened 2023-09-25
Add missing .clone() to model util to fix player model glitches

### LINKED ISSUE #268: Player model fails to render, is incorrectly scaled, or incorrectly oriented

kanyesthaker:
### Description of Changes
Reflections PR neglected to `.clone()` each new instance of the player model to account for the now globally-cached model object. This change makes it so that all player model objects behave independently.

### Link to Issue
closes #268 

### Testing Procedure
Unable to reproduce any of the player model bugs we have noticed after 10 attempts at refreshing, and re-rendering the page, also with pausing and unpausing.


samjgorman (approved):
Great!

Non-blocking question: are we sure we want a 2.5x scaling factor? Seems a bit large to me. Perhaps experiment with 2x?

PR #276, opened 2023-10-01
bug: Fix Cloudflare worker failures by stopping Cloudflare from signing the accept-encoding header

### LINKED ISSUE #275: Unable to download assets from Backblaze with error code 403

kanyesthaker:
### Description of Changes
In the last few weeks our Cloudflare worker fails to successfully fetch assets from Backblaze. This change maps to a recent modification in the worker [template](https://github.com/backblaze-b2-samples/cloudflare-b2/blob/c421e7888ec92ed13fa543a60fe352fb1afd2b8e/index.js#L16) which specifies the HTTPS port, which appears to resolve the issue.

Edit 10/7: Fixes critical issue here https://github.com/backblaze-b2-samples/cloudflare-b2/pull/4

### Link to Issue
closes #275 

### Testing Procedure
- [ ] Each user (Shalin, Sam, Kanyes) should be able to use the world across at least 2 days with 0 download failures


samjgorman:
Why do we need to commit updates to the yarn cache?

storage-api/yarn.lock
samjgorman, line 1: Why are we re-generating the yarn lock?  Doesn't look like we've made any changes to the package.json.

Should we update the version of wrangler in the package.json in this PR? From earlier explorations saw that we were potentially a major version behind. That said, let's avoid committing these yarn lock changes unless they're tied to updates in package.json?

storage-api/worker.ts
```diff
@@ -31,6 +35,8 @@ async function handleRequest(event: any, client: AwsClient) {
   }
 
   const url = new URL(request.url);
+  // url.protocol = HTTPS_PROTOCOL;
```
samjgorman, line 38: 👀

.env.local.sample
```diff
@@ -7,7 +7,7 @@ REACT_APP_TERRAIN_FLAG=false
 REACT_APP_SANDBOX_FLAG=false
 
 # When true, route download requests through Cloudflare CDN. Required when bucket is private.
-REACT_APP_CLOUDFLARE_CDN=false
+REACT_APP_CLOUDFLARE_CDN=true
```
samjgorman, line 10: Is setting this flag to true enough? What if we folded the flag instead?  My understanding is that with this, we're still exposing a public bucket URL in client-side code.  Has this bucket been set back to private? If so, then we likely don't need this flag anymore, and if it is is still public, then likely introducing unneeded security overhead. WDYT?

samjgorman (commented):
Left questions!

storage-api/yarn.lock
kanyesthaker: i think this is because we upgraded the version of yarn to a new major version, and the yarn lock format is different?

storage-api/worker.ts
kanyesthaker: shoot this shouldn't be commented out, good eye

.env.local.sample
kanyesthaker: that's a good point -- let me change the code so that we're not exposing the bucket in the code directly

storage-api/yarn.lock
samjgorman: Got it.  Did some research & think this is correct. On inspection of the yarn.lock we see a different format too.  Looks like we missed updating this in the deps PR.

samjgorman (approved):
Bravo!

PR #277, opened 2023-10-03
Add prettier lint rule for Tailwind

samjgorman:
### Description of Changes
Add officially supported Prettier lint rule for Tailwind CSS, this has become necessary as we've added more CSS to the 2D interface and need to keep CSS readable & consistent

### Testing Procedure
Loaded the world successfully


src/common/constants.ts
```diff
@@ -167,5 +167,7 @@ export abstract class AudioConstants {
 }
 
 export abstract class LightingConstants {
-  static readonly directionalLightPosition: [number, number, number] = [10, 10, 5];
+  static readonly directionalLightPosition: [number, number, number] = [
```
kanyesthaker, line 170: is this necessary?

kanyesthaker (approved):

src/common/constants.ts
samjgorman: Picked up this previously unlinted file while linting, seems like a net positive change so included

PR #284, opened 2023-10-14
Fix "Track has already been sent" client error

samjgorman:
### Description of Changes
Wrap the streamMedia function in a useEffect to ensure that it is only called when dependencies in the useCallback-optimized fn change.  Before, `streamMedia` was firing extraneously as it was called directly in `PlayerDataProvider`, leading us to send streams again that had already been sent

### Testing Procedure
Tested across 2+ peers that on join no `Track has already been sent` error messages observed


src/scene/players/PlayerDataProvider.tsx
```diff
@@ -75,8 +75,11 @@ const PlayerDataProvider = () => {
     onMessage: onMessage,
   });
 
-  streamMedia();
-  log.debug('Instantiated Communicator with onMedia and onMessage callbacks');
+  /**
+   * @effect Stream MediaStream to peers, updating the stream as deps in the useCallback optimized fn change
+   * @deps streamMedia
+   */
+  useEffect(streamMedia, [streamMedia]);
```
kanyesthaker, line 82: neat

src/network/worker.ts
```diff
@@ -273,8 +273,7 @@ const handleSessionPeerMessage = (
   sessionID?: string,
 ) => {
   console.trace(
-    `Received Peer message. Routing from one session to the ${
-      sessionID === activeSessionID ? 'active' : 'initiator'
+    `Received Peer message. Routing from one session to the ${sessionID === activeSessionID ? 'active' : 'initiator'
```
kanyesthaker, line 276: we need to sync our lint rules, i see this happen quite a bit

kanyesthaker (approved):

PR #286, opened 2023-10-16
Add 3-D names for users

kanyesthaker:
### Description of Changes
Users are now able to see the names set by other users in 3-D. These names float above the head of the users in the room, and are visible by all other players at all times. Edits to names happen in real-time.

image


src/scene/players/PeerPlayerName.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { PlayerConstants } from '../../common/constants';
+import { playerNamesMapAtom } from '../../common/playerStore';
+import { PlayerMovementData } from '../../common/types';
+
+type PeerPlayerNameProps = {
+  peerId: string;
+  playerPositionMapRef: MutableRefObject<Map<string, PlayerMovementData>>;
+};
+
+const FONT_URL =
```
kanyesthaker, line 16: I put these here because they're only used in this file. I can also put these in constants.ts/, open to feedback

src/interface/NameInput.tsx
```diff
@@ -36,7 +37,7 @@ const NameInput = () => {
     if (!isPeerConnected) return;
 
     sendData({ type: DataSchemaType.NAME, value: { name } });
-  }, [isPeerConnected]);
+  }, [isPeerConnected, otherPeersInRoom]);
```
samjgorman, line 40: Would be a safer & simpler check to reason about if we make the length of the set a dependency, rather than the set itself. What do you think?

src/scene/players/PeerPlayerName.tsx
samjgorman: This works for now!

src/scene/players/PeerPlayerName.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const playerNamesMap = useAtomValue(playerNamesMapAtom);
+
+  const nameMesh = useRef<Mesh>(new Mesh());
+  const font = useLoader(FontLoader, FONT_URL);
+
+  // Updates the text in the TextGeometry whenever the playerNames atom is modified
+  useEffect(() => {
+    // If no name received, uses the first 4 characters of the Peer ID
+    let name = playerNamesMap.get(peerId);
+    if (!name) name = `User ${peerId.slice(0, 4)}`;
```
samjgorman, line 38: As discussed live, let's refactor this to use planet naming rather than this approach

src/scene/players/PeerPlayerName.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    // Creates a custom TextGeometry
+    // https://threejs.org/docs/#examples/en/geometries/TextGeometry
+    const geometry = new TextGeometry(name, {
+      size: 0.05,
+      height: 0,
```
samjgorman, line 44: Why height 0? Also, why are so many values = 0.005?  Should we lift this into a constant?

src/scene/players/PeerPlayerName.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    // Due to the inverse square law, the size of the name should scale with
+    // the square root of the distance between the two users
+    let scale = Math.sqrt(nextFramePosition.distanceTo(camera.position));
+    nameMesh.current.scale.set(scale, scale, scale);
+
+    // Scale the opacity as the players move closer to each other
+    if (nameMesh.current.material && Array.isArray(nameMesh.current.material)) {
+      nameMesh.current.material.forEach((m) => {
+        m.opacity = Math.min(NAME_OPACITY_MAX, NAME_OPACITY_MIN + 1 / scale);
```
samjgorman, line 102: Why `+1`?

src/scene/players/PeerPlayerName.tsx
```diff
@@ -0,0 +1,110 @@
...
+          position.y + NAME_HEIGHT_ABOVE_MODEL,
+          position.z,
+        ),
+      );
+    nameMesh.current.position.copy(nextFramePosition);
+
+    // Rotate the name so it is always facing the camera
+    nameMesh.current.lookAt(camera.position);
+
+    // Due to the inverse square law, the size of the name should scale with
```
samjgorman, line 94: : ()

src/scene/players/PeerPlayerName.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    // Centers the text so rotations function about the midpoint of the text object
+    nameMesh.current.position.fromArray(PlayerConstants.spawnPosition);
+    geometry.computeBoundingBox();
+    if (geometry.boundingBox) {
+      const centerOffset =
+        -0.5 * (geometry.boundingBox.max.x - geometry.boundingBox.min.x);
+      nameMesh.current.geometry.translate(centerOffset, 0, 0);
+    }
+  }, [playerNamesMap]);
```
samjgorman, line 75: Should we refactor this dep to be just the names entry for this peer, and then fire upon changes in this names entry?  

Seems like this will rather fire on each change of each peer, potentially leading to spurious rerenders or subtle bugs

samjgorman (commented):
Impressive! Left a few questions

src/interface/NameInput.tsx
kanyesthaker: What makes checking the length of the set safer or simpler? Isn't it effectively the same thing?

src/scene/players/PeerPlayerName.tsx
kanyesthaker: "Height" is actually the "depth" of the text. If it's more than 0 the text will not appear "flat". We can lift the 0.005s into a constant, but these values are only ever used here, so I feel like it will  make this less readable

src/scene/players/PeerPlayerName.tsx
kanyesthaker: order of operations! it's NAME_OPACITY_MIN + (1/scale) -- this is to make the opacity scale inversely with distance

src/scene/players/PeerPlayerName.tsx
kanyesthaker: Will this actually trigger? Didn't we have this same issue with A/V, where looking for a particular (key, value) in a Map won't fire as a dependency of a useeffect, because the Map itself is what's stateful, not its attribute?

src/interface/NameInput.tsx
kanyesthaker: resolved offline

src/scene/players/PeerPlayerName.tsx
kanyesthaker: resolved offline

src/scene/players/PeerPlayerName.tsx
shalins: agree with sam can we make it a constant, even if it's used in one spot oftentimes it's just nice to be able to edit all UI affecting things in a single file

samjgorman: We could lift to a file-level constant?

src/scene/players/PeerPlayerName.tsx
```diff
@@ -30,12 +30,15 @@ const PeerPlayerName = ({
 
   const nameMesh = useRef<Mesh>(new Mesh());
   const font = useLoader(FontLoader, FONT_URL);
+  const defaultNameRef = useRef(
+    Names.planets[Math.floor(Math.random() * Names.planets.length)],
```
samjgorman, line 34: Lift into a util now that used 2x?

samjgorman (approved):
Looks good!

PR #288, opened 2023-10-27
Names polish

samjgorman:
## Description of Changes

- Render player's name in 3D above player model in `userSelfModel`
- Redesigned toolbar for pause menu
- Focus & error states for `NameInput`
- Make `PlayerName` more modular by moving data fetching logic out of the component and into the `MediaProvider`. This will make it easier to unify the separate logic across self model & player name components

image

## Testing Checklists
### Basic Author Checklist 
- [x] With 1 user:
- [x] With 2 users:

- [ ] xyz

### Advanced Author Checklist

### Basic Reviewer Checklist


src/interface/UserPlayerSelfModel.tsx
```diff
@@ -86,12 +98,39 @@ const UserPlayerSelfModel = () => {
     PlayerConstants.playerSelfModelPosition,
   );
 
+  // Updates the text in the TextGeometry whenever the playerNames atom is modified
+  useEffect(() => {
+    updateNameMeshRef({
+      name: userName,
+      font,
+      nameMeshRef: nameMesh,
+      size: 0.1,
```
shalins, line 107: magic number: make this a constant in `NamesConstants` please

src/interface/UserPlayerSelfModel.tsx
```diff
@@ -86,12 +98,39 @@ const UserPlayerSelfModel = () => {
...
+    });
+
+    // Slant the name mesh
+    nameMesh.current.geometry.translate(0, 0, 0.5);
```
shalins, line 111: magic number: make this a constant as well

src/interface/UserPlayerSelfModel.tsx
```diff
@@ -86,12 +98,39 @@ const UserPlayerSelfModel = () => {
...
     // Player idle animation
     offset = Math.sin(totalElapsedtime.current) / 20;
     model.current.position.setComponent(1, offset);
     totalElapsedtime.current += delta;
+
+    // Player name animation
+    nextFramePosition.copy(
+      new Vector3(
+        playerVisualMesh.current.position.x,
+        playerVisualMesh.current.position.y + 0.6,
```
shalins, line 128: more magic numbers

src/interface/UserPlayerSelfModel.tsx
```diff
@@ -86,12 +98,39 @@ const UserPlayerSelfModel = () => {
...
+  // Updates the text in the TextGeometry whenever the playerNames atom is modified
+  useEffect(() => {
+    updateNameMeshRef({
+      name: userName,
+      font,
+      nameMeshRef: nameMesh,
+      size: 0.1,
+    });
+
+    // Slant the name mesh
```
shalins, line 110:
```suggestion
    // Rotate the name mesh to have the same orientation as the player model.
```

src/interface/UserPlayerSelfModel.tsx
shalins: also this seems like a bug to me. the previous code for player idle animation is oscillating the y component of the vector by the range of sine (with some scale). the player name, however, seems to be monotonically increasing w.r.t time? wouldn't the text just fly off the screen?

also the units don't make sense to me. are we shifting the y-pos by 0.6m?

also, if we're only updating the y-component of the vector, can we just use the same syntax as the player idle animation for consistency?

src/scene/players/PeerPlayerMediaProvider.tsx
```diff
@@ -96,6 +108,12 @@ const PeerPlayerMediaProvider = ({
     }
   }, [peerStream]);
 
+  const peerName = useMemo(() => {
+    let name = playerNamesMap.get(peerId);
+    if (!name) name = generatePlayerName();
+    return name;
```
shalins, line 114: wait i don't think this is the right behavior, if we don't have a name for a peer yet, i don't think we should be setting it to a random planet, we should show nothing. the random name should only be used for your own name, not other peers. this could lead to subtle bugs where a player is showing up with a random name because of network latency or something.

src/scene/players/PlayerName.tsx
```diff
@@ -0,0 +1,72 @@
...
+  peerId,
+}: PeerPlayerNameProps) => {
+  const { camera } = useThree();
+
+  const nameMesh = useRef<Mesh>(new Mesh());
+  const font = useLoader(FontLoader, NamesConstants.fontUrl);
+
+  // Updates the text in the TextGeometry whenever the playerNames atom is modified
+  useEffect(() => {
+    updateNameMeshRef({ name, font, nameMeshRef: nameMesh });
```
shalins, line 28: thanks for splitting this logic up

src/scene/players/PlayerName.tsx
```diff
@@ -0,0 +1,72 @@
...
+  }, [name]);
+
+  const nextFramePosition = new Vector3(0, PlayerConstants.height, 0);
+
+  useFrame(() => {
+    // If running for a single peer, we don't need to update as position changes
+
+    if (!peerId || !playerPositionMapRef) return;
```
shalins, line 36: also this comment doesn't really make sense to me. do you mean when running for peer players vs user player self model? if so, i think it makes much more sense to send information over as a bool flag in the prop or something than doing this check in a sorta roundabout + unclear way.

src/common/constants.ts
```diff
@@ -67,7 +72,7 @@ export abstract class Colors {
   static readonly bgGrey = 'rgb(220, 220, 220, 0.5)';
 }
 
-export abstract class Names {
+export abstract class NamesConstants {
```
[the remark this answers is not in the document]
samjgorman: Agree, I picked up the `ColorConstants` change and can take on future work to make the rest consistent

src/common/constants.ts
```diff
@@ -67,7 +72,7 @@ export abstract class Colors {
...
   static readonly planets = [
```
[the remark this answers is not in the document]
samjgorman: Got it -- agree that this is more semantic and easier to understand.

src/interface/components/utils/calcClassname.tsx
```diff
@@ -0,0 +1,11 @@
+/**
+ * Combine classnames together, especially useful when 
+ * conditionally adding classnames based on state
+ * @param classes 
+ * @returns 
+ */
+const calcClassnames = (...classes: string[]) => {
```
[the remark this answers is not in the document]
samjgorman: So this is used to construct Tailwind CSS classNames that respond to stateful updates, an example is a button that accepts a disabled prop and needs to conditionally change its CSS based on the `disabled` bool

Ex)

````typescript
<div className=(calcClassnames('h-full', disabled ? 'bg-gray-500': ''}/>
````

Correct that after I removed the disabled check, this is no longer used. I understand that general convention tells us don't add unused code but I do see this as a generally helpful util that's likely to be used in our 2D interface soon.

That said, the friction to re-introducing is low -- so I'll go ahead and remove, and reintroduce in a future PR that needs to use this pattern

File: src/common/utils/names.ts [hunk 11]
```diff
@@ -0,0 +1,58 @@
...
+    new MeshPhongMaterial({
+      color: NamesConstants.innerColor,
+      flatShading: true,
+      transparent: true,
+    }), // Text fill
+    new MeshPhongMaterial({
+      color: NamesConstants.borderColor,
+      flatShading: true,
+      transparent: true,
+    }), // Text border
```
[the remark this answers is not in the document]
samjgorman: Great point, updating

src/interface/UserPlayerSelfModel.tsx
samjgorman: I'm opting to instead define this as a local constant, as it's only used in the `userPlayerSelfModel` code. We can graduate this to `NameConstants` once its reused

src/scene/players/PlayerName.tsx
samjgorman: I think we can remove this code and tackle this problem in a better way if we want to reuse `PlayerName` outside of the `MediaProvider`. I initially wanted to do this but learned that there's actually some separate logic happening in `PlayerName`, such as scaling the name and opacity as players move in 3D space, that are irrelevant to the `userSelfPlayerModel`

src/scene/players/PeerPlayerMediaProvider.tsx
samjgorman: Setting to nothing could lead to a confusing user experience where they view 3D text above their own model from the pause menu but 3D names are not yet visible for other peers.

Can we start by maintaining the behavior above, and then tweak this as we use & QA the world?

src/interface/UserPlayerSelfModel.tsx
samjgorman: Cleaned up the magic numbers!  

My best understanding of what's happening here ( CC @kanyesthaker ) is that on each frame, we are setting the position of the `nameMesh` to be that of the `playerVisualMesh`,  at an additional height above the `playerVisualMesh`.

The `nameMesh` is unaffected by time elapsed, rather only changing when the position of the `playerVisualMesh` changes. For the `userPlayerSelfModel`, this is only when the height of the model is affected by changes in oscillation

Re: what units these are:  believe this is an abstract unit that best correlates to a point in 3D space. See here for docs -- and curious to hear from @kanyesthaker if he can 1. confirm and 2. If this is abstract, discuss tradeoffs for standardizing to meters to unify development in Blender
https://threejs.org/docs/#api/en/math/Vector3

Re: syntax as player idle animation consistency: 
We could set the position directly in such a manner as the idle animation:

````typescript
 nameMesh.current.position.setComponent(
      1,
      playerVisualMesh.current.position.y + SELF_TEXT_HEIGHT_ABOVE_PLAYER,
    );
````

But I don't love the non-obvious behavior we introduce where future devs need to understand that setting component 1 maps to the Y axis. I've refactored to an approach that sets the nextFramePosition directly rather than creating new vectors, which is both more semantically similar to how we handle idle player behavior & more memory efficient while preserving readability.

Ex)

````typescript
 nextFramePosition.set(
        playerVisualMesh.current.position.x,
        playerVisualMesh.current.position.y + SELF_TEXT_HEIGHT_ABOVE_PLAYER,
        playerVisualMesh.current.position.z,
    );

    nameMesh.current.position.copy(nextFramePosition);
````

File: src/interface/components/UserInput.tsx [hunk 12]
```diff
@@ -15,21 +15,21 @@ const UserInput = ({
...
-        placeholder={placeholder}
-        value={value}
-        onFocus={onFocus}
-        onBlur={onBlur}
-      />
-    </div>
+    <input
+      type="text"
+      className={
+        'h-[48px] w-max rounded-md border border-gray-700 bg-black bg-opacity-50 px-4 py-2 text-xs text-white placeholder-white shadow-inner outline-none focus:border-blue-500'
```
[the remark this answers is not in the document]
samjgorman: This is a problem if you need responsiveness on dynamic height / width in your application. Today for pause menu we don't need that this so I'd say this is acceptable. That said, here's a few areas I'd be excited to take our styling moving forward:

1. Introduce a tailwind.config file that describes shared styling primitives that can be used across the app. This will reduce the need for one-off styling like above and introduce more consistency 
2. Within that config, introduce dynamic values based on viewport height / width to help future-proof the interface to future needs to be more responsive

src/common/utils/names.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  // Center name
+  nameMeshRef.current.position.fromArray(
+    PlayerConstants.playerSelfModelPosition,
+  );
+  geometry.computeBoundingBox();
+  if (geometry.boundingBox) {
+    const centerOffset =
+      -0.5 * (geometry.boundingBox.max.x - geometry.boundingBox.min.x);
+    nameMeshRef.current.geometry.translate(centerOffset, 0, 0);
```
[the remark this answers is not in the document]
samjgorman: Agree in principle, tending towards leaving this out of scope for this PR and would prefer @kanyesthaker can organize this code in a way that makes most sense for him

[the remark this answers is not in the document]
shalins: sounds good, i'd prefer if generic operations like this can live inside of a math.ts file (and maybe split apart once that gets too long) so we can pull in as we need. would prefer if it was left as a TODO or something so it doesn't slip through the cracks

src/interface/UserPlayerSelfModel.tsx
```diff
@@ -86,12 +106,37 @@ const UserPlayerSelfModel = () => {
...
+    updateNameMeshRef({
+      name: userName,
+      font,
+      nameMeshRef: nameMesh,
+      size: SELF_MODEL_GEOMETRY_SIZE,
+    });
+
+    // Rotate the name mesh to have the same orientation as the player model.
+    nameMesh.current.geometry.translate(0, 0, SELF_MODEL_Z_SLANT);
+    nameMesh.current.quaternion.copy(angleQuat);
```
shalins, line 120: nit, can we put the comment above the quaternion line instead of the `translate` line? or at least modify it to something like `// Translate first, then rotate the name mesh to have the same orientation as the player model.` so it refers to both?

src/interface/components/UserInput.tsx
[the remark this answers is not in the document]
shalins: great idea! lets revisit this as our styling needs to scale, especially as we introduce new 2d UI

shalins (approved):

src/interface/UserPlayerSelfModel.tsx
shalins: sounds good! I personally prefer setting the component actually but it's not a huge deal and am happy to just move forward with this. also, i had read that wrong initially, which is my mistake, i didn't realize the next mesh was always being set above the _player_ mesh, i read that as nextFramePos = nextFramePos + height.

src/scene/players/PeerPlayerMediaProvider.tsx
shalins: I think that's for the peer player though? so it's setting it for new peers who join who haven't had their name come in yet, not for self peer above their own model?

src/scene/players/PlayerName.tsx
shalins: yeah i get that, but what im suggesting is passing a variable as a prop so you would initialize the props with a flag such as `isPeerPlayer` or something and then in this function you'd read `if (!isPeerPlayer) return;`

src/interface/UserPlayerSelfModel.tsx
shalins: thanks for the detailed explanation btw!

src/interface/components/utils/calcClassname.tsx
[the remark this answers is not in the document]
shalins: ah got it, thanks for the explanation! yeah lets add it back when it's used to avoid confusion

src/scene/players/PeerPlayerMediaProvider.tsx
samjgorman: The self peer reads from the `userName` atom in `playerStore`, where if no name exists in the cache for the atom it generates a new name by calling the same util.

Also, I'd like to give more thought to the related case of after a user has entered a username, and then modifies it back to an empty string -> currently the 2D UI will then show blank screen while 3D peer is shown as random name, and would like to unify this.   I'm going to treat this as non-blocking to this PR getting in and then use the world more to iron out functionality

src/scene/players/PlayerName.tsx
samjgorman: Yeah this is a solid idea, I'll incorporate this into any future work that reuses or decouples `PlayerName` further

PR #289, opened 2023-11-09
server: fix `segfault` and simple-peer audio track crashes

shalins:
## Description of Changes
TLDR: This PR fixes the server crashes when AV is toggled, upgrades our packages, and introduces some additional infra for better debugging on the server side.

**Package Updates:**
- Several libraries used in the project were updated to newer versions. These upgrades were part of debugging the server side crashes, but I've persisted them so we are working off the latest dependencies. Changes have been tested.

**Server Error Handling:**
- The server-side code now includes a segfault handler for creating a 'crash.log' file, which is useful for diagnosing crashes.
 - Global error handling for uncaught exceptions and unhandled promise rejections are now logged.
- `errorhandler` middleware has been added but is configured to only be used in a development environment.

**Development Configuration:**
- A `nodemon.json` file has been created to configure nodemon for the server, enabling an inspect flag set for debugging with Chrome debug tools.

**Server-side Crash Fixes:**
- To fix the server segfaults, two main changes have been added. The first handles the case in which mutiple streams are sent by the same peer, in which case now the older streams are removed before adding the new one. The second is to add try-catch blocks around all simple-peer functions for better error handling when wrtc internal state becomes inconsistent. Previously, these issues would crash the server.
- New enums and interfaces related to media streaming have been added for better media stream management on the server side.

**Client-side Crash Fixes:**
- On the client side, from the `PauseScreen` to the `useUserMediaStream` hook, and all the way to the Communicator API, we now properly handle toggling audio and video on and off without creating new `MediaStream` objects. This allows for simpler control on the client side, fixes the segfault issues on the server, and also fixes the brief "flash" that appears when a peer toggles their AV.

General Cleanups:
- I've added additional comments and tidying up of the code across files that I touched.

## Testing Checklists
### Basic Author Checklist 
- [x] With 1 user:
- [x] With 2 users:

- [x] Tried spamming the AV buttons and that still seems to work properly.

### Advanced Author Checklist
- [x] With 3+ users, the server and client behave in a manner consistent with 2 peers

### Basic Reviewer Checklist

Todo:
- [ ] Issue where on first load, one of the peer's videos does not load.
- [ ] Double check that the useCommunicator hook is being used properly across the client.
- [ ] AV toggles get out of sync with the stream and takes a toggle to reset the state.


server/utils/consts.ts
```diff
@@ -2,7 +2,7 @@
 // https://datatracker.ietf.org/doc/rfc6455/?include_text=1
 export const PingOp = 0x9;
 export const PongOp = 0xa;
-export const DefaultPingInterval = 3000; // milliseconds
+export const DefaultPingInterval = 30000; // milliseconds
```
kanyesthaker, line 5: 30s interval? is this to prevent network-related timeouts?

server/utils/types.ts
```diff
@@ -8,6 +8,18 @@ export interface HeartbeatWebSocket extends WebSocket {
 
 export type TimeoutMap = Map<WebSocket, NodeJS.Timeout>;
 
+export enum StreamType {
```
kanyesthaker, line 11: good thinking

src/network/messageRouter.ts
```diff
@@ -699,9 +703,22 @@ export default class MessageRouter {
...
+    this.mediaStream = stream;
     this.peerManager.sendStream(stream);
   }
 
+  /**
+   * Toggles the playback state of the specified media type tracks within the media stream.
+   * @param {MediaType} type - The type of media to update (audio or video).
+   * @param {boolean} isEnabled - If false, the specified media type is disabled.
+   */
+  updateMediaState(type: MediaType, isEnabled: boolean) {
```
kanyesthaker, line 715: nit: `MediaState` is ambiguous, maybe `updateMediaEnabledState` or `updateTracksEnabled` etc.

src/network/messageRouter.ts
```diff
@@ -37,6 +38,9 @@ export default class MessageRouter {
 
   private initiatorSessionID?: string;
 
+  // Store the media stream object in state for convenient modification.
```
kanyesthaker, line 41:
```suggestion
  // Maintain the media stream object as an instance var for convenient modification.
```

src/network/utils/consts.ts
```diff
@@ -1,7 +1,9 @@
 export const WebSocketURL = 'wss://spacebar-5.herokuapp.com';
```
kanyesthaker, line 1: haven't thought about this too much before, but these constants should probably be in `camelCase` or `UPPER_CASE` -- the constants in the client code are `PascalCase` because they are objects and the constants are their fields. I don't care too much about changing this, but thought I'd point it out to avoid confusing a constant with a class name.

src/network/hooks/useInitializeCommunicator.ts
```diff
@@ -19,7 +18,7 @@ const useInitializeCommunicator = (roomId?: string) => {
 
   if (!roomId) {
     console.warn(
-      `No roomId has been provided, so falling back to default room name ${DEFAULT_ROOM_NAME}`,
+      `No roomId has been provided, so falling back to default room name ${DefaultRoomName}`,
```
kanyesthaker, line 21: see comment on `/consts.ts`

src/network/communicator.ts
```diff
@@ -49,6 +49,15 @@ export default class Communicator {
     this.messageRouter.stream(stream);
   }
 
+  /**
+   * Controls the play/pause state of video or the mute/unmute state of audio.
+   * @param {MediaType} type - The media type to control, audio or video.
+   * @param {boolean} isEnabled - True to play/unmute, false to pause/mute.
+   */
+  toggleMedia(type: MediaType, isEnabled: boolean) {
```
kanyesthaker, line 57: nicely done

src/common/hooks/useUserMediaStream.tsx
```diff
@@ -82,7 +86,11 @@ const useUserMediaStream = () => {
     handleRestartStream();
   };
 
-  return { restartUserMediaStream };
+  const updateUserMediaStream = (mediaType: MediaType, isEnabled: boolean) => {
```
kanyesthaker, line 89: once we've decoupled the communicator instantiation logic from the function export logic, is there any reason to maintain this function (one-liner)?

kanyesthaker (commented):
nice clean code. A couple questions, most don't require code changes

server/app.ts
```diff
@@ -1,12 +1,25 @@
...
 import log from 'loglevel';
+import SegfaultHandler from 'segfault-handler';
 import ws from 'ws';
 
 import { SignalingServer } from './server';
 import { DefaultPingInterval, PongOp } from './utils/consts';
 import { HeartbeatWebSocket, TimeoutMap } from './utils/types';
 
+// Register a handler to create a 'crash.log' file in the event of a crash
+SegfaultHandler.registerHandler('crash.log');
```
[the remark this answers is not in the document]
shalins: yeah good question. it will overwrite older files, no versioning --  but also don't need to be versioned imo, just a single dump is still better than current status quo

server/app.ts
```diff
@@ -83,14 +102,21 @@ server.on('upgrade', (request, socket, head) => {
...
-process.on('exit', () => {
-  clearInterval(pingInterval);
+process.on('SIGTERM', () => {
+  console.log(`Process ${process.pid} received a SIGTERM signal`);
+  process.exit(0);
+});
+
+process.on('SIGINT', () => {
+  console.log(`Process ${process.pid} has been interrupted`);
+  process.exit(0);
```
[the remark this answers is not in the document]
shalins: yeah good point will fix

server/utils/consts.ts
shalins: yes, wanted to bump up by large enough amount to validate whether or not this is causing the issue where peers are disconnected. plus the problem this solves is to prevent the Heroku 55 second idle timeout from shutting down the server, so 30s ping ping round trip still works

src/network/utils/consts.ts
shalins: Yeah that's a good point, I can refactor these in a dedicated future PR

server/server.ts
```diff
@@ -218,14 +227,14 @@ export class SignalingServer {
...
-   * @param {string} peerID (Optional) ID of the peer. If not provided, a new
+   * @param {string} [peerID] ID of the peer. If not provided, a new
    * ID will be generated.
    */
   _createNewPeer(ws: WebSocket, roomID: string, peerID?: string) {
     const peersInRoom = this.rooms.get(roomID).peers;
     const seenPeer = peersInRoom.find((peer) => peer.peerID === peerID);
 
-    log.debug(
+    log.trace(
```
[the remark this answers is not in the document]
shalins: it's very verbose and not debug level imo to log the entire wrtc instance. in my head, debug/info is at sfu + signaling level and trace is more for wrtc and simple peer level logging.

server/server.ts
```diff
@@ -338,14 +347,39 @@ export class SignalingServer {
...
+  _attachStreamToPeer(peer: Peer, senderPeerID: string, stream?: MediaStream) {
     if (stream && peer.isConnectedP2P) {
-      if (!peer.remoteStreamIds?.includes(stream?.id)) {
+      if (
+        peer.remoteStreams?.filter((s) => s.streamID === stream?.id).length == 0
+      ) {
+        // TODO(SHALIN): Rethink logic when we support screen sharing.
+        // If we get a new stream from the same peer, we need to remove
+        // the old stream first.
+        let oldStreamMetadata = peer.remoteStreams?.find(
```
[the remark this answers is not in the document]
shalins: exactly, that's the assumption. eventually we want to leverage the enum i added to allow for one valid stream per media type.

server/server.ts
```diff
@@ -338,14 +347,39 @@ export class SignalingServer {
    * Attaches a stream to a `SimplePeer` instance and keeps track as part
    * of the peer's metadata.
    * @param {Peer} peer Peer to attach the stream to.
+   * @param {string} senderPeerID ID of the peer that originally sent the stream.
    * @param {MediaStream} stream Stream to attach to the peer.
```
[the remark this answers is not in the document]
shalins: This is a function of the stream object arriving asynchronously -> Peer objects can be initialized before the stream arrives, so we make the stream as optional and initialize the peer with no stream, and attach the stream when it arrives. But this means that throughout the code, we treat it as optional.

server/server.ts
```diff
@@ -338,14 +347,39 @@ export class SignalingServer {
...
+      if (
+        peer.remoteStreams?.filter((s) => s.streamID === stream?.id).length == 0
+      ) {
+        // TODO(SHALIN): Rethink logic when we support screen sharing.
+        // If we get a new stream from the same peer, we need to remove
+        // the old stream first.
+        let oldStreamMetadata = peer.remoteStreams?.find(
+          (s) => s.senderPeerID === senderPeerID,
+        );
+        if (oldStreamMetadata && oldStreamMetadata.streamID != stream?.id) {
```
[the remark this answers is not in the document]
shalins: Yeah it is, good catch

server/server.ts
```diff
@@ -357,19 +391,43 @@ export class SignalingServer {
...
     if (stream) {
       this.rooms.get(roomID)?.peers.forEach((p) => {
-        if (p.remoteStreamIds?.includes(stream?.id)) {
-          log.info(`Removing stream ${stream.id} from peer ${p.peerID}`);
-          p.simplePeer?.removeStream(stream);
-          p.remoteStreamIds = p.remoteStreamIds?.filter(
-            (id) => id !== stream?.id,
-          );
+        if (
+          p.remoteStreams?.filter((s) => s.streamID === stream?.id).length > 0
```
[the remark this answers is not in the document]
shalins: No that's not quite the assumption, the assumption is that we can get in a state where we have multiple with _different_ stream ID's but from the same peer, in which case we want to remove the old stream the peer had sent (if any). In this function, the level of abstraction we're working with is just to remove a single stream, so don't think a while loop makes sense.

server/server.ts
```diff
@@ -357,19 +391,43 @@ export class SignalingServer {
...
+   * Removes a stream from a peer and cleans up metadata.
+   *
+   * @param {Peer} peer Peer to remove the stream from.
+   * @param {MediaStream} [stream] Stream to remove.
+   */
+  _removeStreamFromPeer(peer: Peer, stream?: MediaStream) {
+    if (stream && peer.isConnectedP2P) {
+      log.info(`Removing stream ${stream.id} from peer ${peer.peerID}`);
+      try {
+        // TODO(SHALIN): Verify we're not removing the stream from the `senderPeer`,
```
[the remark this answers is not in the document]
shalins: Yeah, I think that makes sense. Thanks for raising. Addressed.

server/server.ts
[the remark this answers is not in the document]
kanyesthaker: crisp explanation, thanks!

kanyesthaker (approved):

PR #290, opened 2023-11-12
physics: frame-independent time integration for physics sim

kanyesthaker:
## Description of Changes
1. Makes collision checking frame-independent. Before, each cycle of the main WebGL renderer would get evenly subdivided into $n$ equally-sized "collision ticks", meaning there would be the same number of ticks per frame but the actual size of the tick would vary. As a result, when the frame was delayed (such as when you navigate between browser tabs), the corresponding ticks would be very large. Since position updates are the integral of velocity with respect to time, this would cause very large position updates (teleporting) when coming back to an AFK (away-from-keyboard) session.
2. Moves major collision logical chunks to a utils file to abstract away any core physics from the main "UserEntity" React component. 
3. Renames variables and core logic to make the underlying code structure more manageable/generalizable

## Testing Checklists
### Basic Author Checklist 
- [x] With 1 user:
- [x] With 2 users:

- [x] Open a SpaceBar tab with 2 peers, and leave the tab for 10 minutes, and then re-enter -- the player should not have moved

### Advanced Author Checklist

### Basic Reviewer Checklist


src/common/constants.ts
```diff
@@ -19,7 +19,7 @@ export abstract class PhysicsConstants {
 
   static readonly pointLight: [number, number, number] = [10, 10, 10];
 
-  static readonly collisionTickRate: number = 5;
+  static readonly physicsTimeStep: number = 1 / 60; // Tick physics simultaions at a minimum of 60FPS
```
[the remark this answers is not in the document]
samjgorman: How did we converge on 60FPS? 

I understand that 120FPS is used in gaming pretty often -- curious if we experiment with a higher check rate if this could provide a better user experience.  

Further -- is there a way to sync the timeStep to the number of frames executed per second in `useFrame`?  Wondering if we can get ourselves into edge case behavior when the collision FPS is out of sync with the FPS of the world

src/common/utils/playerDynamics.ts
```diff
@@ -1,9 +1,35 @@
...
+  collisionDisplacementUnitVector: Vector3;
+  collisionDisplacementVector: Vector3;
+  inputVelocity: Vector3;
+  slidingVelocity: Vector3;
+  mtv: Vector3;
+};
+
+// To avoid re-allocating memory, this object maintains a static
+// collection of references for all complex types used for user physics simulation
+const GlobalCollisionsObjects: GlobalCollisionsType = {
```
samjgorman, line 20: Makes sense to lift this out of the component to avoid re-initializing this object as the component re-renders. Thanks for the comment

src/common/utils/playerDynamics.ts
```diff
@@ -24,15 +50,11 @@ const GlobalControls: GlobalControlsType = {
 };
 
 /**
- * Create a directional vector based on player keyboard inputs
+ * Creates a directional vector based on player keyboard inputs, updates globally-held value
```
samjgorman, line 53: What is the globally-held value? This is unclear

src/common/utils/playerDynamics.ts
```diff
@@ -24,15 +50,11 @@ const GlobalControls: GlobalControlsType = {
...
  * @param params A MovementControlsParams object that tracks player keyboard and mouse controls
- * @returns A unit Vector3 that points in the appropriate direction in player-space
  */
-
-export const getControlsMovementVector = (
-  params: MovementControlsParams,
-  target: Vector3,
-): void => {
+export const saveInputsAsVelocity = (params: InputParams): void => {
```
samjgorman, line 56: Nit: unnecessary to type this as void explicitly, Typescript will correctly infer the return type for this

src/common/utils/playerDynamics.ts
```diff
@@ -54,46 +76,24 @@ export const getControlsMovementVector = (
...
   ]);
 
-  // Get the direction from the keyboard, apply the yaw, and normalize to get a unit vector
-  target
+  // Gets the direction from the keyboard, apply the yaw, and normalize to get a unit vector
+  inputVelocity
     .addVectors(GlobalControls.xInputVelocity, GlobalControls.zInputVelocity)
     .applyQuaternion(GlobalControls.yaw)
-    .normalize();
+    .setLength(PlayerConstants.movementSpeed);
```
samjgorman, line 83: Why are we setting the length?

src/common/utils/playerDynamics.ts
```diff
@@ -54,46 +76,24 @@ export const getControlsMovementVector = (
...
 /**
- * Determine how to adjust the player when colliding with an object
+ * Determines how to adjust the player when colliding with an object
  * @param params PlayerCollisionParams detailing required components to calculate player collision
  * @returns A unit vector pointing in the direction of adjustment if a collision is registered
  */
-export const calculatePlayerCollisions = (
-  params: PlayerCollisionParams,
+export const computeCapsuleBVHCollisions = (
+  { capsule, bvh }: CapsuleBVHParams,
```
samjgorman, line 96: Nice cleanup!

src/common/utils/playerDynamics.ts
```diff
@@ -54,46 +76,24 @@ export const getControlsMovementVector = (
...
 
-type PlayerCollisionParams = {
-  playerCurrentPosition: Vector3;
-  playerRadius: number;
-  playerRay: Line3;
-  playerWorldMatrix: Matrix4;
-  boundsTree?: MeshBVH;
+type CapsuleBVHParams = {
+  capsule: Mesh;
+  bvh?: MeshBVH;
```
samjgorman, line 88: Would prefer explicitness -- dislike abbreviations and it's easier to understand what's going on when it's clear what BVH is

```suggestion
 boundingVolumeHierarchy
?: MeshBVH;
```

src/common/utils/playerDynamics.ts
```diff
@@ -141,9 +141,9 @@ export const calculatePlayerCollisions = (
 
       // If the distance between the player axis and the triangle is smaller
```
samjgorman, line 142: Does this comment need to be updated? Unsure what `the triangle` is referring to

src/common/utils/playerDynamics.ts
```diff
@@ -164,35 +164,74 @@ export const calculatePlayerCollisions = (
...
 
 /**
- * A utility that calculates what the player's "gravity" vector should be
- * on the next frame. This is influenced by what the player is standing on,
- * introducing a concept akin to "friction"
- * @param displacementVector
- * @param currentTickGravity
- * @param isPlayerOnGround
- * @returns
+ * ! (Kanyes) This is a temporary fix for edge-case behavior while
```
samjgorman, line 171: Can you refactor this to a TODO for clarity?

src/common/utils/playerDynamics.ts
```diff
@@ -164,35 +164,74 @@ export const calculatePlayerCollisions = (
...
-  if (!isPlayerOnGround) {
-    displacementVector.normalize();
-    currentTickGravity.y += PhysicsConstants.gravity * tickDelta;
-    target.addVectors(
-      currentTickGravity,
-      displacementVector.multiplyScalar(
-        -displacementVector.dot(currentTickGravity),
-      ),
-    );
+const restrictWithinBoundaries = (position: Vector3) => {
```
samjgorman, line 176:
```suggestion
const setPlayerPositionWithBoundaries = (position: Vector3) => {
```

src/common/utils/playerDynamics.ts
```diff
@@ -1,9 +1,35 @@
...
+type GlobalCollisionsType = {
+  collisionRay: Line3;
+  playerCollisionBoundingBox: Box3;
+  closestTrianglePoint: Vector3;
+  closestPlayerRayPoint: Vector3;
+  collisionDisplacementUnitVector: Vector3;
+  collisionDisplacementVector: Vector3;
+  inputVelocity: Vector3;
+  slidingVelocity: Vector3;
+  mtv: Vector3;
```
samjgorman, line 15: What is this? Can we avoid abbreviations for this

src/common/utils/playerDynamics.ts
```diff
@@ -164,35 +164,74 @@ export const calculatePlayerCollisions = (
...
+  capsule.updateMatrixWorld();
+  mtv.fromArray([0, 0, 0]);
+  computeCapsuleBVHCollisions(
+    {
+      capsule: capsule,
+      bvh: collider.geometry.boundsTree,
+    },
+    mtv,
+  );
+  capsule.position.add(mtv);
```
samjgorman, line 234: See comment re: `mtv` -- makes it quite hard to understand what this line is doing

src/scene/players/UserPlayerEntity.tsx
```diff
@@ -60,51 +56,38 @@ UserPlayerEntityProps) => {
...
 
-  // Instantiate player collision object
-  // The player object is a "capsule", a cylinder with spherical caps at both ends
-  // This geometry lets us emulate both "player height" and makes collision calculations
-  // easier, since we can use radial symmetry.
-  // A capsule is defined by a central axis of revolution and a radius
   useEffect(() => {
-    // Add player displacement to make players not spawn inside each other
-    // HACK (Kanyes, Sam): create a spawn area to prevent this
+    // ! Kanyes: This is a hack until we have a dedicated spawn area
```
samjgorman, line 73: Reformat to TODO.

Out of curiosity -- what is this syntax you're using?

src/scene/players/UserPlayerEntity.tsx
```diff
@@ -113,168 +96,67 @@ UserPlayerEntityProps) => {
...
-  window.addEventListener('blur', () => {
-    log.debug('Movement halted since tab is out of focus');
-    if (playerColliderRef.current?.position) {
-      playerLastSeenPosition.current.copy(playerColliderRef.current.position);
-    }
-  });
+  // Handle the tab going out of focus by pausing the physics simulation.
+  // When the tab goes out of focus, stash the player's current position.
+  // When the tab returns to focus, restor the player's last safe position.
+  // ! (Kanyes) This logic should be moved higher up when we have more simulations running
```
samjgorman, line 112: TODO

src/scene/players/UserPlayerEntity.tsx
```diff
@@ -113,168 +96,67 @@ UserPlayerEntityProps) => {
...
-  // Halt all movement if the tab is not in focus
-  window.addEventListener('blur', () => {
-    log.debug('Movement halted since tab is out of focus');
-    if (playerColliderRef.current?.position) {
-      playerLastSeenPosition.current.copy(playerColliderRef.current.position);
-    }
-  });
+  // Handle the tab going out of focus by pausing the physics simulation.
+  // When the tab goes out of focus, stash the player's current position.
+  // When the tab returns to focus, restor the player's last safe position.
```
samjgorman, line 111:
```suggestion
  // When the tab returns to focus, restore the player's last safe position.
```

src/scene/players/UserPlayerEntity.tsx
samjgorman: Could we add `@effec`t and `@dep` syntax for this `useEffect`?

See codebase for examples, looking to standardize our documentation for useEffects for max clarity

src/scene/players/UserPlayerEntity.tsx
```diff
@@ -113,168 +96,67 @@ UserPlayerEntityProps) => {
...
+      window.removeEventListener('blur', handleBlur);
+      window.removeEventListener('focus', handleFocus);
+    };
+  }, []);
 
   const communicator = useAtomValue(communicatorAtom);
 
-  // Update movement vectors based on camera location
+  // Player physics simulation loop
+  let accumulator = 0;
```
samjgorman, line 138:
```suggestion
  let physicsLoopAccumulator = 0;
```

src/scene/players/UserPlayerEntity.tsx
```diff
@@ -113,168 +96,67 @@ UserPlayerEntityProps) => {
...
-    // Copies the object's position to the camera (moves the camera)
-    camera.position.lerp(playerColliderRef.current.position, 0.5);
 
-    // We have a "tick rate" of the world of 5 collision checks per frame
-    tickDelta = delta / PhysicsConstants.collisionTickRate;
-    for (let i = 0; i < PhysicsConstants.collisionTickRate; i++) {
-      // Grab the camera's current quaternion
+    // Run the collision simulation with frame-independent time integration
+    accumulator += delta;
+    while (accumulator > 0) {
```
samjgorman, line 148: I'm okay with this, but am curious if you've considered other simpler approaches to managing the physics loop state? 

Also, nice work moving these to utils!

src/scene/players/UserPlayerEntity.tsx
samjgorman: Should accumulator be a ref?  If `UserPlayerEntity` rerenders and this value is set back to 0, could this interfere with the physics loop?

samjgorman (commented):
Looking solid! Left some questions, and as discussed async, flagged potential issue where collisions do not apply to bar model

src/common/constants.ts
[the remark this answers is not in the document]
kanyesthaker: we actually don't want the timestep to be sync'ed to the useFrame loop speed. This is how it was working before, which was causing some unexpected behavior, because the actual physics of the world would behave differently based on your hardware. Instead, it's generally better practice to fix the simulation speed, regardless of the actual framerate. 

As to 60 vs 120fps, i don't have a strong opinion. It won't provide any meaningful difference to users, since most if not all users will have their effective frame rate capped by the refresh rate on their screens. 60FPS is pretty standard.

src/common/utils/playerDynamics.ts
kanyesthaker: good to know!

src/common/utils/playerDynamics.ts
kanyesthaker: The movement velocity is a vector pointing in our intended direction, whose magnitude is the player's movement speed (right now a mostly arbitrary constant, though we can ascribe units to it). Previously, we separated "direction" and "magnitude" into two separate computations. Instead, we can use `setLength` directly to save a computation. Minor optimization all things considered.

src/common/utils/playerDynamics.ts
kanyesthaker: I generally agree, but bvh is a pretty standard industry term, and a widely known data structure in the graphics world as well. Plus there's a notion doc explaining its behavior -- i prefer the abbreviation. If you feel strongly I can change it to the lengthier alternative

src/common/utils/playerDynamics.ts
kanyesthaker: refers to the mesh triangle in the mesh that we are intersecting

src/scene/players/UserPlayerEntity.tsx
kanyesthaker: https://gafferongames.com/post/fix_your_timestep/

src/scene/players/UserPlayerEntity.tsx
kanyesthaker: sure! good idea

src/scene/players/UserPlayerEntity.tsx
kanyesthaker: resolved offline

src/common/utils/playerDynamics.ts
kanyesthaker: whoops, forgot to comment. MTV stands for `minimum translation vector`, also a common concept in the collision detection literature https://fulmanski.pl/zajecia/tippgk/zajecia_20162017/wyklad_cwiczenia_moje/collision.pdf

This one is more obscure, so i'm fine expanding it to its full form. I just generally find it easier to read math code when I'm looking at "variables" instead of "concepts"

src/common/utils/playerDynamics.ts
samjgorman: Let's keep `bvh` then -- if this becomes a common point of confusion in the code we'll change later

src/scene/players/UserPlayerEntity.tsx
samjgorman: This was a really helpful read -- thoughts on linking to this article in the PR description, or updating our documentation to describe how our fixed timestep works with this linked?

samjgorman (approved):
Approving to unblock, nice work! Be sure to include the revisions we converged on & this is good to go!

PR #292, opened 2023-11-15
Fix: Zero state names polish

samjgorman:
## Description of Changes

-Update the default player names
- Refactor how we read & write to userName:  now the player is strictly responsible for assigning a default value to itself and sending this to peers.  This is simpler than our prior approach to have the player also set default names for other peers
- Switch from writing to `localStorage` directly to using `useAtomWithStorage`
- Change UX such that when a player enters an empty string in the `NameInput`, a default value is assigned, rendered in selfModel, and broadcast to peers

image

## Testing Checklists
### Basic Author Checklist 
- [ ] With 1 user:
  - [ ] The world loads, the user can move/look around, and physics behave properly
  - [ ] User can pause their client, mute/unmute themselves, and change their name
- [ ] With 2 users:
  - [ ] A second user can join the same room, and both users can see each others model/video and hear each others audio
  - [ ] State changes (name changes, A/V toggles) propagate between peers as expected

- [ ] Verified that the player's name in SelfModel is set to a default when the string in NameInput is empty
- [ ] Verified that when string in NameInput is empty, the same default name shown in SelfModel is broadcast to peers and rendered for other peers

### Advanced Author Checklist

### Basic Reviewer Checklist


src/interface/NameInput.tsx
```diff
@@ -1,24 +1,35 @@
...
   const otherPeersInRoom = useAtomValue(otherPeersInRoomAtom);
+  const isPeerConnected = useAtomValue(isConnectedP2PAtom);
   const setIsPlayerTyping = useSetAtom(isPlayerTypingAtom);
-  const { sendData } = useCommunicator();
   const [name, setName] = useAtom(userNameAtom);
-
-  const isPeerConnected = useAtomValue(isConnectedP2PAtom);
+  const [nameInUserInput, setNameInUserInput] = useState<string>(name);
 
   const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
```
shalins, line 19: Can write this function more simply without code duplication:
```
  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
    // If the user removes the input name,
    // generate a new default & send to the server
    let name = e.target.value ? generatePlayerName() : e.target.value; // <-- I changed to check for null and undefined also

    setName(name);
    setNameInUserInput(e.target.value ? '' : e.target.value);
    sendData({ type: DataSchemaType.NAME, value: { name: name } });
  };
```

src/interface/NameInput.tsx
```diff
@@ -28,14 +39,18 @@ const NameInput = () => {
...
     <UserInput
-      value={name}
-      placeholder={name}
+      value={nameInUserInput}
+      placeholder={nameInUserInput || 'Enter a name'}
+      // Value passed in is hydrated by a value from localStorage,
+      // which means that on first render SSR & client have different values
+      // This doesn't cause an issue, but we do suppress the warning below
+      // https://nextjs.org/docs/messages/react-hydration-error
+      suppressHydrationWarning
```
shalins, line 53: can we actually move this comment to where it's used instead of the parent component? so put in the `UserInput.tsx` file.

src/interface/UserPlayerSelfModel.tsx
```diff
@@ -65,17 +65,6 @@ const UserPlayerSelfModel = () => {
     `Added peer video mesh at ${playerVisualMesh.current.position.toArray()}`,
   );
 
-  /**
```
shalins, line 68: nice cleanup

src/interface/NameInput.tsx
```diff
@@ -28,14 +39,18 @@ const NameInput = () => {
...
   return (
     <UserInput
-      value={name}
-      placeholder={name}
+      value={nameInUserInput}
+      placeholder={nameInUserInput || 'Enter a name'}
+      // Value passed in is hydrated by a value from localStorage,
+      // which means that on first render SSR & client have different values
+      // This doesn't cause an issue, but we do suppress the warning below
+      // https://nextjs.org/docs/messages/react-hydration-error
```
shalins, line 52: Also instead of suppressing the warning, why can't we just use solution one or two? I feel like disabling SSR is a clearer way to indicate the intention than passing this suppression prop. My suggestions come from the same link you suggested: https://nextjs.org/docs/messages/react-hydration-error

src/scene/players/PeerPlayerMediaProvider.tsx
```diff
@@ -108,11 +101,7 @@ const PeerPlayerMediaProvider = ({
     }
   }, [peerStream]);
 
-  const peerName = useMemo(() => {
-    let name = playerNamesMap.get(peerId);
-    if (!name) name = generatePlayerName();
-    return name;
-  }, [playerNamesMap.get(peerId)]);
```
shalins, line 115: Do we really need to remove the useMemo here? It will now do the "get" on each render, it's not expensive but just wondering.

shalins (approved):

src/scene/players/PeerPlayerMediaProvider.tsx
```diff
@@ -43,7 +36,7 @@ const isAudioAvailable = (peerStream: MediaStream) =>
   peerStream?.getAudioTracks()?.[0]?.readyState === MediaStreamReadyState.Live;
 
 /**
- * This component handles retrieving and appropriately handling
+ * This component handles retrieving and handling
```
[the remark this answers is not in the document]
samjgorman: Felt like noise so removed!

src/common/playerStore.ts
```diff
@@ -25,12 +25,14 @@ export const isVideoEnabledAtom = atom<boolean>(true);
...
 
 export const userMediaStreamErrorAtom = atom<string | null>(null);
 
-export const userNameAtom = atom<string>( generatePlayerName());
-
 export const isPausedAtom = atom<boolean>(false);
 
 export const isPlayerTypingAtom = atom<boolean>(false);
+
+export let userNameAtom = atom<string>(generatePlayerName());
```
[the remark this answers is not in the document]
samjgorman: yup, good catch here

src/scene/players/PeerPlayerMediaProvider.tsx
samjgorman: Yup, we should remove -- our guidance for when to memoize is to only use this to optimize expensive calculations, where this `get` should be cheap.  So we can remove and simplify the code

src/interface/NameInput.tsx
samjgorman: I've changed this to now dynamically import the `UserInput` component with SSR off! Agree that this is a better solution, thanks for double-checking!

src/interface/NameInput.tsx
samjgorman: No longer relevant as we dynamically import instead of suppress the warning now

src/interface/NameInput.tsx
samjgorman: Refactored for clarity! Went with a similar approach as above

PR #293, opened 2023-11-18
Feat: Rooms from URL

samjgorman:
## Description of Changes

- Support custom rooms from specifying a roomId from the room path of the URL. Ex) localhost:3000/`sam`
- Establish dynamic route on `/[roomId]`, and read the passed in roomId from the url. Then, pass this as a prop to `App` to initialize a communicator room.

## Testing Checklists
### Basic Author Checklist 
- [x] With 1 user:
- [x] With 2 users:

- [x] Asserted that entering text after the root path constructs a room that multiple peers can connect on

### Advanced Author Checklist

### Basic Reviewer Checklist


src/network/hooks/useInitializeCommunicator.ts
```diff
@@ -1,35 +1,27 @@
...
+import { useSetAtom } from 'jotai';
 import log from 'loglevel';
 import { useEffect } from 'react';
 
 import Communicator from '../communicator';
-import { communicatorAtom, isCommunicatorInitializedAtom } from '../store';
-import { DefaultRoomName } from '../utils/consts';
+import { communicatorAtom } from '../store';
+
+type UseInitializeCommunicatorProps = {
```
kanyesthaker, line 8: is this more opaque than having the props be directly used as arguments?

kanyesthaker: we haven't needed to add a prop to this since it was first written, so not sure how future-proof we need to be, no strong opinion

src/pages/[roomId].tsx
```diff
@@ -0,0 +1,16 @@
+import { useRouter } from 'next/router';
```
Comment (kanyesthaker) [inline src/pages/[roomId].tsx:1]: I couldn't tell why this component's name was in brackets, so I found this https://nextjs.org/docs/pages/building-your-application/routing/dynamic-routes

src/pages/[roomId].tsx
```diff
@@ -0,0 +1,16 @@
...
+import React from 'react';
+
+import App from '../App';
+
+const Room = () => {
```
Comment (kanyesthaker) [inline src/pages/[roomId].tsx:6]: I'm a little confused how this works. As far as I can see we don't actually use the Room component anywhere, and it's unintuitive to me to have a component return `<App/>`, which is usually the top-most level component. Could you either (a) add a link in a docstring to the appropriate documentation explaining how this works, or (b) update the README for the repo so that I know how to use this file if I need to update it in the future?

kanyesthaker (approved):

src/pages/[roomId].tsx
Comment (kanyesthaker) [reply src/pages/[roomId].tsx:6]: ok, after reading through the nextjs docs I do see how this works. I still think the documentation changes are warranted

Comment (samjgorman) [reply src/pages/[roomId].tsx:6]: Added a docstring in [roomId].tsx! Agree that this isn't very clear unless you're familiar with the NextJS documentation.  As we add multiple dynamic URLs we can then refactor to part of readme.  For more in-depth understanding refer to our next JS migration guide that I wrote recently in our internal docs

src/network/hooks/useInitializeCommunicator.ts
samjgorman: We could get away with passing this directly as an argument, but I prefer props as typed objects universally as...

1. Props become self-documenting, as the object key documents what the prop represents
2. Scales better as components evolve to have multiple props

src/pages/[roomId].tsx
Comment (samjgorman) [reply src/pages/[roomId].tsx:1]: It's likely in brackets because the next/router import has multiple properties compared to just `useRouter`, so we're destructuring the import to just grab the `useRouter` hook

PR #295, opened 2023-12-03
communicator: updated hook to only set communicator callbacks once

shalins:
## Description of Changes
Updated `useCommunicator` hook to only call the callback functions once by:
* Removing the previous logic of setting the callbacks on an empty useEffect (which would override the previously set functions each time we used the hook in one of the components.
* Adding a new initializeCallbacks function that must now explicitly be called in order to set the callbacks for communicator.
* Logging an error if any of the components attempts to set a callback that has already been set by one of the other components. 

## Testing Checklists
### Basic Author Checklist 
- [x] With 1 user:
- [x] With 2 users:

- [x] The new callbacks are set correctly and only once instead of being overwritten.
- [x] Attempting to set the callbacks more than one results in a runtime error.

### Advanced Author Checklist
- [x] With 2 users, the server remains healthy after an extended amount of time in the world
- [x] With 3+ users, the server and client behave in a manner consistent with 2 peers

### Basic Reviewer Checklist


yarn.lock
kanyesthaker, line 1: i don't mind package upgrades in general, but you intentionally wanted to bump all our packages?

src/network/hooks/useCommunicator.ts
```diff
@@ -21,107 +37,121 @@ interface UseCommunicatorProps {
...
-    });
-
-    communicator?.on('media', (stream, peerID) => {
-      log.debug(`Received stream from peer ${peerID}`);
-      if (onMedia) {
-        onMedia(stream, peerID);
-      }
-    });
+  const eventMapping: {
+    [key: string]: {
```
kanyesthaker, line 53: a) can we make this a type?
b) instead of `key` maybe `callbackName`?

src/network/hooks/useCommunicator.ts
```diff
@@ -21,107 +37,121 @@ interface UseCommunicatorProps {
...
+      eventKey: string;
```
kanyesthaker, line 54: maybe we can make the callback names/strings an enum to be a little stricter with typing. Also makes understanding this a little easier

src/network/hooks/useCommunicator.ts
```diff
@@ -21,107 +37,121 @@ interface UseCommunicatorProps {
...
+              log.debug(
+                `Event ${mappedEventKey} triggered with args: ${JSON.stringify(
+                  args,
+                )}`,
+              );
+              if (callback) {
+                callback(...args);
+              }
+            });
+            atomSetter(callback !== undefined || callback !== null);
```
kanyesthaker, line 146: here i think you can just use `callback != null` and type coersion will handle both cases

src/network/hooks/useCommunicator.ts
```diff
@@ -21,107 +37,121 @@ interface UseCommunicatorProps {
...
+          } else {
+            log.error(`Cannot set \`${eventKey}\` callback more than once.`);
+          }
+        },
+      );
+    },
+    [communicator, handlePeerEvent, eventMapping],
```
kanyesthaker, line 153: does including eventmapping as a dependency do anything?

src/network/utils/consts.ts
```diff
@@ -23,3 +23,12 @@ export const MaxCachedMessages = 10000;
 export const StreamPeerCheckInterval = 100; // milliseconds
 
 export const BroadcastChannelName = 'BROADCAST_CHANNEL';
+
+// Event keys for the Communicator.
+export const OnConnectEvent = 'connect';
```
kanyesthaker, line 28: see above, consider making these an enum

src/network/hooks/useCommunicator.ts
```diff
@@ -21,107 +37,121 @@ interface UseCommunicatorProps {
...
+        ([eventKey, { eventKey: mappedEventKey, atomValue, atomSetter }]) => {
+          const callback = callbacks[eventKey as keyof CommunicatorCallbacks];
+
+          if (!atomValue) {
+            communicator?.on(mappedEventKey, (...args) => {
+              if (eventKey === 'onPeerJoined' || eventKey === 'onPeerLeft') {
+                const [data, peerID] = args;
+                handlePeerEvent(eventKey, data, peerID);
+              }
+              log.debug(
```
kanyesthaker, line 137: do we get log spam from the `onMessage` callback? maybe downgrade to trace?

src/network/hooks/useCommunicator.ts
```diff
@@ -21,107 +37,121 @@ interface UseCommunicatorProps {
...
-  }, [communicator?.isConnectedWS]);
+    },
+    [setOtherPeersInRoom],
+  );
+
+  const initializeCallbacks = useCallback(
+    (callbacks: CommunicatorCallbacks) => {
+      Object.entries(eventMapping).forEach(
+        ([eventKey, { eventKey: mappedEventKey, atomValue, atomSetter }]) => {
+          const callback = callbacks[eventKey as keyof CommunicatorCallbacks];
```
kanyesthaker, line 129: This method also means that the callbacks used to initialize communicator must be named the correct callback names, so if i wanted to initialize `onMedia` with a callback `handlePlayerStreams` I can't (not 100% sure this is accurate). How can we (a) either make it so that the consumer of this API knows what the callbacks can be named or (b) introduce some more flexibility?

src/scene/players/PlayerDataProvider.tsx
```diff
@@ -70,10 +70,14 @@ const PlayerDataProvider = () => {
     }
   };
 
-  const { streamMedia } = useCommunicator({
-    onMedia: onMedia,
-    onMessage: onMessage,
-  });
+  const { initializeCallbacks, streamMedia } = useCommunicator();
```
kanyesthaker, line 73: see previous comments re: strictness of naming

src/network/hooks/useCommunicator.ts
```diff
@@ -21,107 +37,121 @@ interface UseCommunicatorProps {
...
+
+          if (!atomValue) {
+            communicator?.on(mappedEventKey, (...args) => {
+              if (eventKey === 'onPeerJoined' || eventKey === 'onPeerLeft') {
```
kanyesthaker, line 133: consts? tbh i'm a little confused by mapping from callback to event to string names. there is `onPeerJoined` which is an event name, and also `peerJoined` which is the constant corresponding to `OnPeerJoinedEvent`. Can we clean this up?

kanyesthaker (commented):
a couple comments, lmk when you've had a chance to scan

src/network/utils/consts.ts
```diff
@@ -24,11 +24,24 @@ export const StreamPeerCheckInterval = 100; // milliseconds
...
-export const OnConnectEvent = 'connect';
-export const OnMessageEvent = 'message';
-export const OnMediaEvent = 'media';
-export const OnDisconnectEvent = 'disconnect';
-export const OnErrorEvent = 'error';
-export const OnPeerJoinedEvent = 'peerJoined';
-export const OnPeerLeftEvent = 'peerLeft';
+// Communicator event listener keys.
+export enum CommunicatorEventKey {
+  OnConnect = 'connect',
```
kanyesthaker, line 29: nit: I think it's a little odd that these enums have the same name for their variants, but different values. Don't care that much though.

src/network/hooks/useCommunicator.ts
kanyesthaker: Just checked out your branch with `debug` log level and the `onMessage` spam makes the logs unusable (screenshot) -- i agree that we shouldn't set something this high level to `trace` by default. Let's either exclude this log message, or add special handling for `onMessage` in particular (the only "spammy" event).

image

kanyesthaker (approved):
approving to unblock, nicely done; but pls see the log spam issue

src/network/hooks/useCommunicator.ts
shalins: ok good point -- that this would be annoying, i'll demote to trace

src/network/utils/consts.ts
shalins: Hmm, I don't think it's odd! They're namespaced differently, even if the variants are the same -- seems quite idiomatic to me.

PR #300, opened 2023-12-05
physics: fix collision regressions from collisions refactor

kanyesthaker:
## Description of Changes
This PR fixes a few physics regressions from #290 . 
1. Caps the minimum frame rate so that we do not accumulate more frames than we can reasonably process.
2. Moves the processing of player inputs outside of the actual simulation loop to prevent undesired movement in between frames
3. Adds an extra condition when updating the "sliding" velocity of the player to ensure that the player can never slip through the floor
4. Removes hacky bounding box logic that would hide more critical errors in collisions code
5. Adds logging to catch future issues with collisions or BVH generation

## Testing Checklists
If your change adds critical functionality that should be tested on every PR, edit ``pull_request_template.md`` and add it to the checklist.
### Basic Author Checklist 
- [x] With 1 user:
- [x] With 2 users:

- [x] Reload the world 20 times with no collisions bugs

### Advanced Author Checklist

### Basic Reviewer Checklist


src/common/constants.ts
```diff
@@ -79,7 +81,7 @@ export abstract class NamesConstants {
     'Hawking',
     'Feynman',
     'Curie',
-    'Armstrong'
+    'Armstrong',
```
kanyesthaker, line 84: @samjgorman do you have trailing commas off in your lint settings?

src/common/utils/playerDynamics.ts
```diff
@@ -27,6 +29,7 @@ const GlobalCollisionsObjects: GlobalCollisionsType = {
   inputVelocity: new Vector3(),
   slidingVelocity: new Vector3(),
   mtv: new Vector3(),
+  loggingVector: new Vector3(),
```
shalins, line 32: seems to be unused

src/common/utils/playerDynamics.ts
```diff
@@ -167,41 +170,18 @@ export const computeCapsuleBVHCollisions = (
...
+ * @param position The position vector of the player's model
  */
-const updateSlidingVelocity = () => {
+const updateSlidingVelocity = (position: Vector3) => {
   const { slidingVelocity, mtv } = GlobalCollisionsObjects;
-  const shouldStopSliding =
+  const hasCollisionsFromGround =
     mtv.y >
     Math.abs(PhysicsConstants.physicsTimeStep * slidingVelocity.y * 0.25);
+  const isOnGround = position.y < PlayerConstants.height;
```
shalins, line 183: this conditional should also check that position.y is > 0 right? currently if player is stuck inside the ground for whatever reason, this would think it's on ground

src/common/utils/playerDynamics.ts
```diff
@@ -167,41 +170,18 @@ export const computeCapsuleBVHCollisions = (
...
  * This vector determines the tangent to a sloped surface so that objects will "slide" down slopes
+ * @param position The position vector of the player's model
  */
-const updateSlidingVelocity = () => {
+const updateSlidingVelocity = (position: Vector3) => {
   const { slidingVelocity, mtv } = GlobalCollisionsObjects;
-  const shouldStopSliding =
+  const hasCollisionsFromGround =
     mtv.y >
     Math.abs(PhysicsConstants.physicsTimeStep * slidingVelocity.y * 0.25);
```
shalins, line 182: out of curiousity, what's the 0.25 factor applied to the mtv check? if this is referring to the kinematic formula for translation delta, shouldn't the factor be 0.5? would be nice to make this a constant for clarity and/or add comment.

src/common/constants.ts
```diff
@@ -19,6 +19,8 @@ export abstract class PhysicsConstants {
 
   static readonly pointLight: [number, number, number] = [10, 10, 10];
 
+  static readonly maxAccumulatedFrames = 5;
```
shalins, line 22:
```suggestion
  static readonly maxAllowedAccumulatedFrames = 5;
```

src/scene/players/UserPlayerEntity.tsx
```diff
@@ -136,26 +147,40 @@ UserPlayerEntityProps) => {
...
-      camera.getWorldQuaternion(cameraAngle);
-      camera.position.copy(playerCapsuleRef.current.position);
-      saveInputsAsVelocity({
-        isMovingLeft: isMovingLeft,
-        isMovingRight: isMovingRight,
-        isMovingForward: isMovingForward,
-        isMovingBackward: isMovingBackward,
-        angle: cameraAngle,
-      });
+    // If we're running slower than 12FPS, it's ok to drop frames
```
shalins, line 178: this comment can get out of date very quickly if 5 is changed to 6. see comment on adding a combined constant and adding the comment there instead of in here.

src/common/constants.ts
```diff
@@ -19,6 +19,8 @@ export abstract class PhysicsConstants {
...
+
```
shalins, line 23: add the comment and constant in this file:
```
// If the physics accumulator is running slower than 12FPS, it's ok to drop frames.
static readonly physicsFrameRateThreshold = PhysicsConstants.maxAllowedAccumulatedFrames * PhysicsConstants.physicsTimeStep
```

shalins (commented):
Left some comments, but approving to unblock. Also tested the world on your branch by refreshing over 30 times under various conditions (refresh -> switch to new tab -> come back; refresh -> move to bg for 30s -> come back; refresh and stay on tab; etc.) Was not able to reproduce any issues

src/common/utils/playerDynamics.ts
kanyesthaker: ah good eye

src/common/utils/playerDynamics.ts
kanyesthaker: they're equivalent since the feet of the player being at 0 is the same as the head of the player being at PlayerConstants.height, since the position vector corresponds to the player's head. All this check does is make sure that if, for some reason, we fail a collision check on a frame for any reason, the player does not start falling below the plane

src/scene/players/UserPlayerEntity.tsx
kanyesthaker: good point, this is an oversight

src/common/utils/playerDynamics.ts
shalins: I was talking about the displacement formula: https://byjus.com/physics/kinematics-equations/

But yeah, that sounds good, thanks for walking me through it

src/common/utils/playerDynamics.ts
shalins: Ok got it. Was confusing because here (https://github.com/IrisHub/SpaceBar/blob/be7cf50ec35aad1c87528068dd25e8c2ce12963a/src/common/utils/playerDynamics.ts#L238) there's a check for capsule being below zero.

shalins (approved):

src/common/utils/playerDynamics.ts
kanyesthaker: ah got it, yeah that's confusing. The second check is to make sure that _after_ all the physics updates, the player is in the correct position for the next render tick. I can rename this so it's more clear

PR #301, opened 2023-12-10
[Infra] Introduce custom domain & upgrade to HTTPS from NextJS middleware

samjgorman:
## Description of Changes

- Introduce NextJS middleware that for domains within an allowlist, upgrades the request from HTTP to HTTPS.  
- [Outside of PR] Configured DNS targets from Heroku to Namecheap domain of `spacebar.build`. See internal doc for config & instructions https://www.notion.so/irisapp/Custom-domain-configuration-d002d61277304f2f8cb3f7bbb83dabeb?pvs=4
- Bonus: fix a blocking deployment issue of `Error: Cannot find module 'errorhandler' for the server` by moving package from `devDepedencies` to `dependencies`

## Testing Checklists
### Basic Author Checklist 
- [x] With 1 user:
- [x] With 2 users:

- [x]  Loaded spacebar.build and www.spacebar.build multiple times across rooms and peers to assert that can upgrade to HTTPS and behave as expected
- [x] Can deploy server to Heroku and operates as expected

### Advanced Author Checklist

### Basic Reviewer Checklist


src/middleware.ts
```diff
@@ -0,0 +1,26 @@
...
+  if (
+    currentEnv === 'production' &&
+    req.headers.get('x-forwarded-proto') !== 'https' &&
+    domainsToUpgrade.includes(req.headers.get('host') ?? '')
+  ) {
+    return NextResponse.redirect(
+      `https://${req.headers.get('host')}${req.nextUrl.pathname}`,
+      301,
+    );
+  }
```
shalins, line 22: can we explicitly check that the host is not nil (it shouldn't be but still) instead of falling back to empty string?
```suggestion
  const currentHost = req.headers.get('host');
  if (
    currentEnv === 'production' &&
    req.headers.get('x-forwarded-proto') !== 'https' &&
    currentHost &&
    domainsToUpgrade.includes(currentHost)
  ) {
    return NextResponse.redirect(
      `https://${currentHost}${req.nextUrl.pathname}`,
      301,
    );
  }
```

src/middleware.ts
```diff
@@ -0,0 +1,26 @@
...
+  return NextResponse.next();
```
shalins, line 23: nit: can we also move `middleware.ts` to be inside of `./src/network/` instead of just in `./src`?

shalins (commented):
Approving to unblock, nice work!

shalins (approved):

src/middleware.ts
samjgorman: Done!

PR #302, opened 2023-12-10
AV: fix AV state messaging

kanyesthaker:
## Description of Changes
The previous A/V fix tried to send the updated A/V from state in the same useEffect as the corresponding setState invocation. React does not update state immediately but rather asynchronously. A React state variable will never update in the same render cycle. https://react.dev/learn/state-as-a-snapshot

This fix:
1. Moves the "sendData" call to the appropriate component, `PauseScreen`
2. Logs the result on the client end appropriately

In the below video, note that the state logged to console is accurate. Also note that the "video flickering" result is back, which is what we'd expect currently.

https://github.com/IrisHub/SpaceBar/assets/31911175/514fc0f2-13cc-4b8d-b831-b9a68135a12f

## Testing Checklists
### Basic Author Checklist 
- [x] With 1 user:
- [x] With 2 users:

- [x] See included video above

### Advanced Author Checklist

### Basic Reviewer Checklist


src/interface/PauseScreen.tsx
```diff
@@ -26,6 +28,14 @@ const PauseScreen = () => {
   const isVideoEnabled = useAtomValue(isVideoEnabledAtom);
   const isAudioEnabled = useAtomValue(isAudioEnabledAtom);
   const { updateUserMediaStream } = useUserMediaStream();
+  const { sendData } = useCommunicator();
+
+  useEffect(() => {
```
shalins, line 33: instead of moving this out of the useUserMediaStream hook, can we keep in there and use the `isEnabled: boolean` argument instead of the jotai atom for sending state? i'd like pausescreen to be self contained as a UI component, without needing to also house networking logic.

src/interface/PauseScreen.tsx
```diff
@@ -16,6 +16,8 @@ import {
 } from '../common/constants';
 import useUserMediaStream from '../common/hooks/useUserMediaStream';
 import { isAudioEnabledAtom, isVideoEnabledAtom } from '../common/playerStore';
+import { DataSchemaType } from '../common/types';
+import useCommunicator from '../network/hooks/useCommunicator';
```
shalins, line 20: see comment on moving this back

src/scene/players/UserPlayerEntity.tsx
```diff
@@ -67,7 +67,6 @@ UserPlayerEntityProps) => {
       new MeshStandardMaterial(),
     ),
   );
-  log.debug('Instantiated player collider ref');
```
shalins, line 70: 👀

src/common/hooks/useUserMediaStream.tsx
```diff
@@ -23,9 +22,9 @@ const AVErrorMessages: Record<string, string> = {
 };
 
 const useUserMediaStream = () => {
-  const { toggleMedia, sendData } = useCommunicator();
-  const [isVideoEnabled, setVideoEnabled] = useAtom(isVideoEnabledAtom);
-  const [isAudioEnabled, setAudioEnabled] = useAtom(isAudioEnabledAtom);
+  const { toggleMedia } = useCommunicator();
+  const setVideoEnabled = useSetAtom(isVideoEnabledAtom);
+  const setAudioEnabled = useSetAtom(isAudioEnabledAtom);
```
shalins, line 27: see comment about moving back

src/scene/players/UserPlayerEntity.tsx
kanyesthaker: yeah this was a useless log, it's not accurate and spams output

src/interface/PauseScreen.tsx
kanyesthaker: fair enough

src/common/hooks/useUserMediaStream.tsx
```diff
@@ -98,9 +98,17 @@ const useUserMediaStream = () => {
     switch (mediaType) {
       case MediaType.Audio:
         setAudioEnabled(isEnabled);
+        sendData({
+          type: DataSchemaType.MEDIA_METADATA,
+          value: { isAudioEnabled: isEnabled, isVideoEnabled: isVideoEnabled },
+        });
```
shalins, line 104: nit: instead of duplicating code here, can we please move the sendData function outside the switch and do something like this:
```
        sendData({
          type: DataSchemaType.MEDIA_METADATA,
          value: {
              isAudioEnabled: mediaType == MediaType.Audio ? isEnabled : isAudioEnabled, 
              isVideoEnabled: mediaType == MediaType.Video ? isEnabled : isVideoEnabled 
          },
        });
```

src/common/hooks/useUserMediaStream.tsx
```diff
@@ -111,7 +111,12 @@ const useUserMediaStream = () => {
     toggleMedia(mediaType, isEnabled);
     sendData({
       type: DataSchemaType.MEDIA_METADATA,
-      value: { isAudioEnabled: isAudioEnabled, isVideoEnabled: isVideoEnabled },
+      value: {
+        isAudioEnabled:
+          mediaType == MediaType.Audio ? isEnabled : isAudioEnabled,
```
shalins, line 116: ok very sorry but one last thing -> can we call this bool `updatedIsEnabled`? i think it would read more clearly.

shalins (approved):
approved but left one last (optional) comment

PR #305, opened 2023-12-12
graphics: fix ephemeral spawning bug

kanyesthaker:
## Description of Changes
Fixes an ephemeral bug where the player would appear below the surface of the bar for 1 frame before the physics ticker begins

The issue is that the player's physics are paused before the player fully loads. When they're paused the player has a position of (0, 0, 0) (the default). The camera only snaps to the player's position once the physics ticker starts ticking, which is delayed by a few frames. By syncing the saved vector, the player's position, and the camera position on load, we prevent any desync between the three.

## Testing Checklists
### Basic Author Checklist 
- [x] With 1 user:
- [x] With 2 users:

- [x] Refreshed 10 times to ensure no flickering, with multiple peers

### Advanced Author Checklist

### Basic Reviewer Checklist


src/scene/players/UserPlayerEntity.tsx
```diff
@@ -69,16 +69,22 @@ UserPlayerEntityProps) => {
...
     // ! Kanyes: This is a hack until we have a dedicated spawn area
     const minX = -10; // furthest left the player will spawn
     const maxX = 10; // furthest right the player will spawn
     const xDisplacement = Math.floor(Math.random() * (maxX - minX + 1)) + minX;
-    playerCapsuleRef.current.position.copy(
-      new Vector3(...PlayerConstants.spawnPosition).add(
-        new Vector3(xDisplacement, 0, 0),
-      ),
+    let spawnPosition = new Vector3(...PlayerConstants.spawnPosition).add(
+      new Vector3(xDisplacement, 0, 0),
```
shalins, line 83: to make future reviews faster, could you perhaps highlight where the actual bug was fixed? where was the issue? what changed? helps build context for reviewers so we can catch it in future reviews :)

in this case, im assuming that the bug was caused by not setting the `savedPositionRef` -- is that correct? why did that cause the player to spawn below the surface?

shalins, line 83: this is out of curiosity, why are we setting just the `xDisplacement`?

shalins (approved):

src/scene/players/UserPlayerEntity.tsx
kanyesthaker: this is a hack until we get a good spawn area -- the little rug in front of the bar extends along the x-axis.

kanyesthaker: my bad -- The issue is that the player's physics are paused before the player fully loads. When they're paused the player has a position of (0, 0, 0) (the default). The camera only snaps to the player's position once the physics ticker starts ticking, which is delayed by a few frames. By syncing the saved vector, the player's position, and the camera position on load, we prevent any desync between the three. Updated PR description

shalins: great, thanks for clarifying and for pushing the fix!

PR #313, opened 2024-01-08
feat(rooms): Create & leave rooms from UI with root domain 

samjgorman:
## Description of Changes

UI layer changes:
- Add a `Root` page on the base route with a UI that allows users to create rooms (opted for the end user terminology of spaces for clarity)
-Add a custom font, used in a `Logo` component to render a wordmark in a header on the root page 
- Added ability to leave a room from the rooms UI, and modified the design of the room name button 
- Create a smart default room name on loading the `Root` page, which can then be edited by users. There are approx 8100 combinations (9 adjectives x 9 nouns x 100 digits), which means that probability of a collision is roughly 0.0001 percent.  This is more than sufficient at our early stage to provide a programmatic safeguard against collisions -- although more work remains to prevent collisions where users input custom room names
- Code refactor for the `Interface`directory to explicitly separate out code according to major features (HUD, pause screen, Home page)

Frontend data fetching changes:

- Explicitly destroy the peer connection before triggering a hard refresh upon leaving a room from the UI.  This adds additional safety that the connection for this peer is closed on leaving a room
- Wipe frontend global state upon leaving a room from the UI.  This is functionally accomplished by writing to the browser window object, which triggers a hard refresh and wipes any Jotai global state stored in memory. This is instead of using Next's default router, which pushes a new route to the window while preserving state.  

Communicator changes:
- Expose SimplePeer's `peer.destroy()` method across Communicator code, such that it is accessible to the frontend client to call to explicitly close the given peer connection

Root UI: 

image

Updated UI with ability to exit a space and refreshed design for copy to clipboard
image

## Testing Checklists
### Basic Author Checklist 
- [x] With 1 user:
- [x] With 2 users:

See this Notion guide with a description of how this feature was tested, given the scope of the proposed changes: 
https://www.notion.so/irisapp/Test-guide-for-Rooms-c530ad137a014e4984463660d96b5c0f?pvs=4

### Advanced Author Checklist

### Basic Reviewer Checklist


src/interface/components/Button.tsx
```diff
@@ -2,7 +2,9 @@ import React, { ReactNode } from 'react';
 
 import { combineClasses } from '../../common/utils/interface';
 
-export type ButtonWidth = 'small' | 'medium' | 'large';
+export type ButtonWidth = 'small' | 'medium' | 'large' | 'x-large';
+
+export type ButtonTheme = 'primary' | 'secondary';
```
kanyesthaker, line 7: Do we have a way to keep these kinds of styles in sync? like a set of Figma components etc.? Not saying we need a full fledged design system, but it's hard for me to understand how to add interface components without knowing the definitions for

src/interface/components/Button.tsx
```diff
@@ -2,7 +2,9 @@ import React, { ReactNode } from 'react';
 
 import { combineClasses } from '../../common/utils/interface';
 
-export type ButtonWidth = 'small' | 'medium' | 'large';
+export type ButtonWidth = 'small' | 'medium' | 'large' | 'x-large';
```
kanyesthaker, line 5: could we use an enum instead of `|`-chaining strings as a type?

src/interface/components/Button.tsx
```diff
@@ -21,7 +24,7 @@ type ButtonWithText = BaseButtonProps & {
 
 type ButtonWithIcon = BaseButtonProps & {
   icon: ReactNode;
-  text?: string;
+  text?: string | ReactNode;
```
kanyesthaker, line 27: this type isn't obvious to me

src/interface/components/UserInput.tsx
```diff
@@ -23,7 +23,7 @@ const UserInput = ({
     <input
```
kanyesthaker, line 23: out of scope for this PR, but it's quite difficult to track iterative changes to tailwind

src/pages/_app.tsx
```diff
@@ -1,10 +1,18 @@
 import '../styles/globals.css';
 
 import type { AppProps } from 'next/app';
+import localFont from 'next/font/local';
 import React from 'react';
 
+const displayFont = localFont({
+  src: '../../public/fonts/jgs.woff',
```
kanyesthaker, line 8: thoughts on downloading the font from b2 vs. putting it in `/public` and bundling?

src/common/utils/rooms.ts
```diff
@@ -0,0 +1,19 @@
...
+    RoomsConstants.defaultRoomAdjectives[
+      Math.floor(Math.random() * RoomsConstants.defaultRoomAdjectives.length)
+    ];
+  const roomNoun =
+    RoomsConstants.defaultRoomNouns[
+      Math.floor(Math.random() * RoomsConstants.defaultRoomNouns.length)
+    ];
+
+    // generate a random number between 1 and 100
+  const randomInt = Math.ceil(Math.random() * 100);
```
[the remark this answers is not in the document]
samjgorman: Yup!  Worked out a similar conclusion in the PR description, glad we're on the same page

src/interface/components/Button.tsx
samjgorman: Initial rationale against was that I didn't want to add extra imports for consuming components (ex `ButtonWidth.small`)

But given that a user would need to step into the component and inspect the string types rather than getting type hinting before & after, as they do if this was typed as an `enum`, I'm okay with this tradeoff

src/interface/components/Button.tsx
samjgorman: Open to discussing this live -- we do have some reusable Figma components that are pretty close to what was implemented: https://www.figma.com/file/XaxiNefH6CPfmo2auSXqVg/SpaceBar?type=design&node-id=1973%3A30&mode=design&t=A4JNXuenXhE9Kkvm-1

I think we should take a lightweight approach to this in the early days as our brand & design system evolved organically via experimentation -- and then converge once patterns are more established

src/pages/_app.tsx
samjgorman: I'll abstract that question slightly to: Should we host our fonts locally or download from an external source?

There's traction in the community for hosting locally: 
https://news.ycombinator.com/item?id=25300396

Our font size today is <50kb, so it seems like the network RTT of making a request to B2 for such a small asset may not justify shifting over to remote storage. 

I'm open to revisiting this if our fonts start becoming significant in size -- otherwise I'd likely want to maintain local hosted for iteration speed & potentially perf

src/interface/components/UserInput.tsx
samjgorman: What makes this difficult? Not having a visual diff to compare against perhaps? Agree that a lot of tailwind code, when separated from the visual context in which it was written, can be difficult to visualize

src/pages/_app.tsx
kanyesthaker: that's fair, i don't think the concern is RTT so much as a failed request which would mess up page rendering. Let's keep it hosted locally

src/common/utils/rooms.ts
```diff
@@ -0,0 +1,19 @@
+import { RoomsConstants } from '../constants';
+
+const generateRoomName = () => {
```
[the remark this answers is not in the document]
samjgorman: Good idea!

src/interface/Header.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const Header = () => {
```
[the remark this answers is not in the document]
samjgorman: Good idea! Here's the implemented refactor -- WDYT?: 

1. Move components that were strictly used for specific components out of the general `components` folder and into feature specific folders
2. Split into `HUD`, `Home`, and `Pause` folders

image

src/interface/components/Button.tsx
samjgorman: In this case, this was from an experiment I hadn't cleaned up! 

For future reference though, typing as `ReactNode` means that JSX can be passed to that prop. So in this case believe I was prototyping passing a styled `div` as text instead of just a `string`

src/interface/Header.tsx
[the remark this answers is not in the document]
kanyesthaker: yep, i'm good with this for now. I think we can do a more pure refactor separating pages from components, but this is fine

src/common/utils/math.ts
```diff
@@ -121,3 +121,6 @@ export const boxBlur = (array2D: FlatArray2D) => {
     }
   }
 };
+
+export const sampleFromList = (list: any[]) =>
```
kanyesthaker, line 125: can we use typescript generics for the typing here so we don't have to keep casting? i.e.
`export const sampleFromList<T>(list: T[]): T`

src/interface/components/UserInput.tsx
kanyesthaker: yes, also tailwind symbols aren't that interpretable. feels the same as adjusting numerical constants

src/common/constants.ts
```diff
@@ -147,7 +173,7 @@ export abstract class SkyboxImages {
 
   // This represents the skybox in the /city skybox directory
   // It's titled 5 because this is selected from across 5 alternates
-  // contained in the /city dir
+  // contained in the /city directory
```
kanyesthaker, line 176: nit, only if you want -- i think we should add a //TODO(Shalin) here to clean this up. P3 at best.

src/interface/HUD/HeadsUpDisplay.tsx
kanyesthaker, line 1: I like the convention that the root component for a directory ahs the same name as the subdirectory, so `HUD/HUD.tsx` is my preference. I'll be refactoring a good chunk of the client code to abide by this convention in the next day or two

kanyesthaker (commented):
leaving some more comments, will approve after i've gotten a chance to test on my own machine

src/interface/components/UserInput.tsx
samjgorman: This gets better once you spend more time reading Tailwind but definitely agree that the maze of abbreviations can be hard to skim without pretty strong context

src/common/constants.ts
```diff
@@ -79,6 +79,32 @@ export abstract class ColorsConstants {
   static readonly bgGrey = 'rgb(220, 220, 220, 0.5)';
 }
 
+export abstract class RoomsConstants {
+  static readonly defaultRoomAdjectives = [
```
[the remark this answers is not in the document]
samjgorman: I removed longer words like algorithmic and networked and replaced with <= 6 letter words -- we can continue to shorten and iterate from there after this PR gets in

src/interface/Home/Home.tsx
```diff
@@ -0,0 +1,46 @@
...
+import generateRoomName from '../../common/utils/rooms';
+import Button, { ButtonTheme, ButtonWidth } from '../components/Button';
+import UserInput from '../components/UserInput';
+import Header from './Header';
+
+const Home = () => {
+  const [newRoomName, setNewRoomName] = useState<string>(generateRoomName());
+
+  const handleJoinRoom = () => {
+    // It's important that we trigger a hard refresh rather than
```
samjgorman, line 13: We technically do not need to trigger a hard refresh on going to the new room name from the home UI, as we've already cleared Jotai state from a hard refresh on leaving the room -- but this approach is likely more robust to future changes if we modify global state from the home page, and allows us to assume a clean slate for global state upon entering a new room

src/common/utils/math.ts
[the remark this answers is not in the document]
kanyesthaker: sure, that's fine for now

kanyesthaker: Sure, that's fine. But i don't think generics are prematurely optimizing, they're an intentional and powerful part of the language that we should be comfortable exploiting :)

src/common/utils/navigation.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Trigger a hard refresh of the page by changing the window location
+ * as an alternative to using Next's router
+ *
+ * This is useful for when you want to trigger a hard refresh and clear
+ * application state
+ *
+ * @param path - Path to redirect to
+ */
+export const pushHardRefresh = (path: string) => (window.location.href = path);
```
kanyesthaker, line 10: out of scope: i don't know if this needs to be its own utility? and if you feel like it does, it should be in `src/interface` and not `src/common`.

src/pages/_app.tsx
kanyesthaker: nit: I think react projects look for the /public directory by default as HOME, so you can just do `/public/fonts/jgs.woff`.

Also, since launch is on the horizon, let's just make sure that we take care of any accreditation for things like fonts (all our 3-D models are fully Creative Commons, not sure abt this font)

src/pages/index.tsx
```diff
@@ -44,8 +44,8 @@ Sentry.init({
   },
 });
 
-const Root = () => {
-  return <App />;
+const HomeContainer = () => {
```
kanyesthaker, line 47: Feels a little unintuitive to have `HomeContainer` in `/pages/index.tsx`

kanyesthaker (approved):
lgtm, a couple minor nits but approving. Can you add a test step or two to the PR template so that we check this part of the product on every PR? Since it's our only real guaranteed testing system

src/common/utils/navigation.ts
samjgorman: I mainly wanted to name as a utility so that I could make it obvious that the functionality was to push a hard refresh, as otherwise writing to the window href directly isn't quite clear that it takes on this important functionality. It's also used twice, which combined was enough for me to split into a util. I'm good moving to `interface`

src/pages/_app.tsx
samjgorman: Good check!  This font is from Velvetyne, an open source type foundry. See their TOS here -- we're clear to use how we see fit!

https://velvetyne.fr/about/

samjgorman: Unfortunately I get a not found error from setting from the /public dir, but appreciate the suggestion

kanyesthaker: sorry, just /fonts not /public/fonts

PR #316, opened 2024-01-19
Client-side Refactor Part I: Players

kanyesthaker:
## Description of Changes
Substantial refactor of Players code. This refactor has the following purposes:
1. **Simplify code** by reducing code duplication, increasing encapsulation, and targeting separation of concerns
2. **Reduce overabstraction** by bringing functions and constants down from global context into minimal local scope
3. **Increase readability** by deepening the file tree, adding docstrings, and renaming poorly named variables, components, and functions

## Testing Checklists
### Basic Author Checklist 
- [x] With 1 user:
- [x] With 2 users:

- [ ] xyz

### Advanced Author Checklist

### Basic Reviewer Checklist


src/common/constants.ts
```diff
@@ -18,25 +18,7 @@ export abstract class PhysicsConstants {
   static readonly gravity: number = -30;
 
   static readonly pointLight: [number, number, number] = [10, 10, 10];
-
-  // If we're running slower than 12FPS, it's ok to drop frames
-  static readonly maxAllowedAccumulatedFrames = 5;
```
samjgorman, line 23: I'm glad we're refactoring global constants to folder or file level constants -- agree this has become increasingly sprawling & harder to reason about as the project has grown

src/scene/players/hooks/useCustomPlayerFrames.tsx
```diff
@@ -0,0 +1,58 @@
+import { useFrame } from '@react-three/fiber';
+import { MutableRefObject } from 'react';
+import { Object3D, Quaternion, Vector3 } from 'three';
+
+import { PlayerMovementData } from '../../../common/types';
```
samjgorman, line 5: OOS: One cleanup item I've been excited about is introducing aliases for imports such that we can move away from relative imports.  Have found that beyond 2 directories relative imports become unwieldy.  For ex) PlayerMovementData can instead be imported from something like @scene/common/types or @data/common/types

src/common/hooks/useCustomFrames.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { MutableRefObject } from 'react';
+import { Object3D, Quaternion, Vector3 } from 'three';
+
+import { Animation } from '../constants';
+
+const useCustomFrames = () => {
+  /**
+   * Static useFrame that clamps an object's properties on each frame.
+   * Useful if your animations expect a static origin on each tick
+   * @param objectRef The object being reset to a static state
```
samjgorman, line 11: I find that param-level docstrings tend to not be very informative -- my philosophy has shifted to only documenting at the param-level when there is special logic or non-obvious behavior that I can't ascertain from the name + type of the variable

src/common/hooks/useCustomFrames.tsx
```diff
@@ -0,0 +1,62 @@
+import { useFrame } from '@react-three/fiber';
+import { MutableRefObject } from 'react';
+import { Object3D, Quaternion, Vector3 } from 'three';
+
+import { Animation } from '../constants';
+
+const useCustomFrames = () => {
```
samjgorman, line 7: Given that the two hooks within this do not contain shared state, it may be simpler to split these into two independent `useFrame` hooks.  Up to you

src/scene/players/MyPlayer.tsx
samjgorman, line 1: `MyPlayer` feels a bit awkward as a name -- I didn't mind the distinction between `User` and `Peer`

src/scene/players/Players.tsx
```diff
@@ -21,14 +21,14 @@ import {
 } from '../../common/types';
 import useCommunicator from '../../network/hooks/useCommunicator';
 import { otherPeersInRoomAtom } from '../../network/networkStore';
-import PeerPlayerMediaProvider from './PeerPlayerMediaProvider';
-import UserPlayerEntity from './UserPlayerEntity';
+import MyPlayer from './MyPlayer';
+import PeerPlayer from './peerPlayer/PeerPlayer';
 
 /**
- * The PlayerDataProvider defines common data elements that are to be passed
```
samjgorman, line 28: Given that `Provider` has a very specific meaning in React, and that this file did not implement a context provider, def on board with the name change. I do question whether `Players` is descriptive enough to ascertain what this file does conceptually

src/scene/players/hooks/useCustomPlayerFrames.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { SPAWN_POSITION } from '../constants';
+
+const useCustomPlayerFrames = () => {
```
samjgorman, line 8: Consider earlier comment whether it makes sense to split into separate hooks -- whatever we opt for, will need consistency

src/scene/players/peerPlayer/Collider.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { SPAWN_POSITION } from '../constants';
+import useCustomPlayerFrames from '../hooks/useCustomPlayerFrames';
+
+type ColliderProps = {
+  peerId: string;
+  playerPositionMapRef: MutableRefObject<Map<string, PlayerMovementData>>;
+  playerCollisionMapRef: MutableRefObject<Map<string, Mesh>>;
+};
+
+const Collider = ({
```
samjgorman, line 16: Appreciate how straightforward this file is to read, nice work!

samjgorman (approved):
Goodbye tech debt -- it _is_ goodbye, and _not_ see you later! 👋🏻

src/scene/players/hooks/useCustomPlayerFrames.tsx
kanyesthaker: yes, i agree as well. I think we should do this once this refactor is done (i.e. cleaning up the rest of `src/`) so we know exactly how common these chained relative imports are -- i have a hunch that 80% of the time we don't need to be importing from `/common` anyway

src/common/hooks/useCustomFrames.tsx
kanyesthaker: I just like having more information on hover in vscode :)

src/common/hooks/useCustomFrames.tsx
kanyesthaker: they don't, but instead of maintaining an entire `frames` directory inside of the `hooks` directory within each component, i think it's simpler to do it this way. Also (while not used right now) these frames could conceivably want access to shared Threejs state like the camera, scene graph, or renderer, in which case it'd make sense to put them together anyway. For now I think i'll keep them clumped together, unless this file gets too unwieldy or disparate.

src/scene/players/MyPlayer.tsx
kanyesthaker: This was @shalins suggestion -- I'm not too tied to it, but i do think that "User" is an overloaded term (since we talk about "our users" separately from the "user player")

src/scene/players/Players.tsx
kanyesthaker: Ultimately I think a file like this should be named after what it renders vs. what its effects do, and this file does simply render all the players in the room

PR #320, opened 2024-01-25
improve(rooms): Support for joining rooms by pressing enter from UserInput keyDown prop

samjgorman:
## Description of Changes

- Add support for passing a `keyDown` handler for `UserInput`
- Use this in `Home` to support entering a room by pressing enter when the input UI is focused

## Testing Checklists
If your change adds critical functionality that should be tested on every PR, edit ``pull_request_template.md`` and add it to the checklist.
### Basic Author Checklist 
- [x] With 1 user:
  - [x] User can pause their client, mute/unmute themselves, and change their name, with the appropriate interactions showing in the pause menu
- [x] With 2 users:
  - [x] Can leave a room from the UI, rejoin that room from the UI, then leave & conect from the UI to a new room

- [ ] xyz

### Advanced Author Checklist

### Basic Reviewer Checklist


src/interface/Home/Home.tsx
```diff
@@ -29,6 +35,7 @@ const Home = () => {
               placeholder="Name of your space"
               maxLength={30}
```
shalins, line 36: const?

src/interface/Home/Home.tsx
```diff
@@ -15,6 +15,12 @@ const Home = () => {
     pushHardRefresh(`/${newRoomName}`);
   };
 
+  const handleKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === 'Enter') {
```
shalins, line 19: const? also do certain keyboard events ever get registered as lowercase `enter`?

shalins (approved):
Approving to unblock but noticed some magic numbers in the code

src/interface/Home/Home.tsx
samjgorman: Will verify but from our work writing `usePlayerControls` hook believe they're coded as uppercase

src/interface/Home/Home.tsx
samjgorman: Good idea

src/interface/Home/Home.tsx
samjgorman: See here: https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_code_values

shalins: nice! thanks for the reference, as a nit for the future, would be great to include a map going from enum => string with a comment including that link above, so future keypresses could just be queried as `KeyboardInput[KeyboardInput.Enter]` for example

samjgorman: This is a good idea for future work!  We can then have `usePlayerControls` and this logic read from the same shared enum. As a first step towards this I'll link to this documentation in the code

samjgorman: Actually decided to just bundle in this refactor in this PR -- code should be simpler + more maintainable now

